PR #4, opened 2022-07-12
Simple samples

### LINKED ISSUE #8: Sample request: mTLS hello world

cretz:
## What was changed

Added some basic feature samples

## Why?

Java has these and in Go we've needed them because they are great to point people to. Docs often don't do as good of a job as code does at demonstrating code features to coders.

Closes #8


hello/hello_async_activity_completion.py
```diff
@@ -0,0 +1,88 @@
...
+        await handle.complete(f"{input.greeting}, {input.name}!")
+
+
+@workflow.defn
+class GreetingWorkflow:
+    @workflow.run
+    async def run(self, name: str) -> str:
+        # Use execute_activity_method which lets us reference a method instead
+        # of a function
+        return await workflow.execute_activity_method(
```
cretz, line 54: This is waiting on https://github.com/temporalio/sdk-python/pull/69

hello/hello_activity_choice.py
```diff
@@ -0,0 +1,121 @@
...
+                    )
+                )
+            else:
+                raise ValueError(f"Unrecognized fruit: {item.fruit}")
+        return "".join(ordered)
+
+
+async def main():
+    # Start client
+    client = await Client.connect("http://localhost:7233")
```
mfateev, line 91: Why HTTP? Is gRPC using HTTP endpoints?

hello/hello_activity.py
```diff
@@ -1,27 +1,36 @@
...
+from dataclasses import dataclass
 from datetime import timedelta
 
 from temporalio import activity, workflow
 from temporalio.client import Client
 from temporalio.worker import Worker
 
 
+@dataclass
+class ComposeGreetingInput:
```
bergundy, line 12: You should probably document why the sample uses a dataclass and not multiple activity params

hello/hello_activity_choice.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@activity.defn
+async def order_oranges(amount: int) -> str:
+    return f"Ordered {amount} Oranges..."
+
+
+@dataclass
+class ShoppingListItem:
+    fruit: str
```
bergundy, line 35: Why not enum?

hello/hello_activity_choice.py
bergundy: This is how core expects the connection string (as a URL).
The `http` protocol can be replaced with `smtp` and it still works which is not great.
I'd recommend changing this to be just address or at least verifying on the core side that the url is an HTTP(s) url.
Note that an `https` URL will cause core to connect using tls.

hello/hello_async_activity_completion.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@dataclass
+class ComposeGreetingInput:
+    greeting: str
+    name: str
+
+
+class GreetingComposer:
+    def __init__(self, client: Client) -> None:
```
bergundy, line 17: Might as well put the client on activity context, would be super useful.
I want to do this in TS too. Even though the worker uses a different connection type the relevant connection options are similar.

hello/hello_cancellation.py
```diff
@@ -0,0 +1,84 @@
...
+    @workflow.run
+    async def run(self) -> None:
+        # Execute the forever running activity, and do a cleanup activity when
+        # it is complete (on error or cancel)
+        try:
+            await workflow.execute_activity(
+                never_complete_activity,
+                start_to_close_timeout=timedelta(seconds=1000),
+                # Always set a heartbeat timeout for long-running activities
+                heartbeat_timeout=timedelta(seconds=2),
```
bergundy, line 41: NOTE that this isn't a strict requirement, it just helps the worker configure throttling

bergundy: I'd still recommend setting it, the above comment is just a note

hello/hello_child_workflow.py
```diff
@@ -0,0 +1,57 @@
...
+class ComposeGreeting:
+    @workflow.run
+    async def run(self, input: ComposeGreetingInput) -> str:
+        return f"{input.greeting}, {input.name}!"
+
+
+@workflow.defn
+class GreetingWorkflow:
+    @workflow.run
+    async def run(self, name: str) -> str:
```
bergundy, line 25: No dataclass??
Kidding, I think we should show that those aren't strictly required.
Same way we don't bother with data classes for return values either.

hello/hello_child_workflow.py
```diff
@@ -0,0 +1,57 @@
...
+
+async def main():
+    # Start client
+    client = await Client.connect("http://localhost:7233")
+
+    # Run a worker for the workflow
+    async with Worker(
+        client,
+        task_queue="hello-child-workflow-task-queue",
+        workflows=[GreetingWorkflow, ComposeGreeting],
```
bergundy, line 41: Looking at all of this registration logic and I'm pretty sure Python users will find this annoying.
So many Python frameworks out there collect just by using the decorator.
We should consider adding this ability to the SDK IMHO.

hello/hello_continue_as_new.py
```diff
@@ -0,0 +1,83 @@
...
+            ComposeGreetingInput("Hello", name),
+            start_to_close_timeout=timedelta(seconds=10),
+        )
+
+        # Continue as new every 2 seconds until we are told to complete. This is
+        # just a demonstration. In production code you might want to continue as
+        # new when the history is growing too large. Continue as new is
+        # essentially a workflow restart (though another workflow can be used).
+        try:
+            await workflow.wait_condition(lambda: self.exit_requested, timeout=2)
```
bergundy, line 40: Hmm.. timeout should be a timedelta or the are name should have a unit.

hello/hello_continue_as_new.py
```diff
@@ -0,0 +1,83 @@
...
+        result = await workflow.execute_activity(
+            compose_greeting,
+            ComposeGreetingInput("Hello", name),
+            start_to_close_timeout=timedelta(seconds=10),
+        )
+
+        # Continue as new every 2 seconds until we are told to complete. This is
+        # just a demonstration. In production code you might want to continue as
+        # new when the history is growing too large. Continue as new is
+        # essentially a workflow restart (though another workflow can be used).
```
bergundy, line 38: I don't think this is a good sample for continue as new, we should show actual code that needs it instead

hello/hello_cron.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        print(f"Result: {result}")
+
+
+async def main():
+    # Start client
+    client = await Client.connect("http://localhost:7233")
+
+    # Run a worker for the workflow
+    async with Worker(
```
bergundy, line 39: What happens if a user hits ctrl-c after the worker has started?
Does the worker gracefully shutdown?

hello/hello_local_activity.py
```diff
@@ -0,0 +1,57 @@
+import asyncio
```
bergundy, line 1: In TS we marked local activities as experimental since they require some more time in production before we consider them stable.
Python is still considered alpha so I wouldn't strictly bother to mention this but since we're pretty confident about the rest of the features I'd consider mentioning.

hello/hello_mtls.py
```diff
@@ -0,0 +1,96 @@
...
+        return await workflow.execute_activity(
+            compose_greeting,
+            ComposeGreetingInput("Hello", name),
+            start_to_close_timeout=timedelta(seconds=10),
+        )
+
+
+async def main():
+    # Load certs from CLI args
+    parser = argparse.ArgumentParser(description="Use mTLS with server")
```
bergundy, line 38: Probably need server name override too

hello/hello_parallel_activity.py
```diff
@@ -0,0 +1,65 @@
...
+async def say_hello_activity(name: str) -> str:
+    return f"Hello, {name}!"
+
+
+@workflow.defn
+class SayHelloWorkflow:
+    @workflow.run
+    async def run(self) -> List[str]:
+        # Run 5 activities at the same time
+        results = await asyncio.gather(
```
bergundy, line 20: I'd mention `as_completed` as an alternative

hello/hello_signal.py
```diff
@@ -0,0 +1,72 @@
...
+            )
+
+            # Drain and process queue
+            while not self._pending_greetings.empty():
+                greetings.append(f"Hello, {self._pending_greetings.get_nowait()}")
+
+        return greetings
+
+    @workflow.signal
+    async def submit_greeting(self, name: str) -> None:
```
bergundy, line 32: Doesn't need to be async right?
I'm considering disallowing async signal handler in TS altogether because users don't understand that they need to wait for all handlers before completing their workflow and end up losing progress.

It's much easier to reason about workflow code if all "io" is done in the main workflow thread.

bergundy (approved):
A lot of my comments here are not blocking for this PR or not relevant for actual samples, just random thoughts from seeing how the SDK is used.

I think we could use a sample showing how to run sync activities (ones that run in an executor).

Also cancelling from workflow should be covered eventually, I assume you're aware, also not blocking this PR.

hello/hello_activity_choice.py
cretz: > I'd recommend changing this to be just address or at least verifying on the core side that the url is an HTTP(s) url.

Can you clarify the recommendation? Are you saying in Python we should change to just address or in Core? If the former, are you saying I should prepend the scheme? How do I know whether it should be `http` or `https`? I can't use the presence of TLS config because there may be no config needed (unless I made users use an empty one to use TLS which is a bit ugly).

Most gRPC implementations require that you opt in to TLS separate from the address. And addresses accept custom schemes (see https://github.com/grpc/grpc/blob/master/doc/naming.md and `HostPort` at https://pkg.go.dev/go.temporal.io/sdk/internal#ClientOptions). I can change to anything we want, I just need to know what we want.

hello/hello_activity_choice.py
cretz: I don't think it adds anything when everything is right next to each other. Might as well use string consts. But I can add enums I guess.

hello/hello_activity.py
cretz: I will do it here. I hope you are not suggesting I have to do this everywhere I use a data class forever.

hello/hello_cancellation.py
cretz: Yup. I tell people to always set it as a best practice.

hello/hello_child_workflow.py
cretz: Yeah, I'm being a bit practical in several places. The dataclass-everywhere, while recommended in production, just isn't how quick scripters will do things (and lots of Python people seem to be scripters).

hello/hello_continue_as_new.py
cretz: This is built to match https://docs.python.org/3/library/asyncio-task.html#asyncio.wait. Really, none of the Python APIs use `timedelta`s (https://docs.python.org/3/library/asyncio-task.html#asyncio.sleep, https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_later, etc).

I am trying to avoid accepting `Union[float, timedelta]` everywhere to avoid confusion, so I figured following other `asyncio` APIs on just this call makes sense. But I can change if wanted. (I wish we had reviewed these before)

hello/hello_cron.py
cretz: The worker does get gracefully shutdown because of `async with`. `Worker.__aexit__` is invoked. If a user didn't use `async with` but instead `Worker.run()`, they'd be responsible for `Worker.shutdown()` in an `except` or `finally`.

hello/hello_local_activity.py
cretz: Not following this comment. Are you saying I should mark local activities in Python experimental? Why do they require more time in production than any other feature of the Python SDK?

hello/hello_signal.py
cretz: It does need to be async because I use `await` for putting on the queue. I could use `put_nowait` since the queue does not have a max size, but this is clear too.

> It's much easier to reason about workflow code if all "io" is done in the main workflow thread.

A caller choice IMO.

cretz:
> I think we could use a sample showing how to run sync activities (ones that run in an executor).

I can add this. Technically there about 100 things I could think of.

> Also cancelling from workflow should be covered eventually, I assume you're aware, also not blocking this PR.

Can you clarify what you mean by cancelling from workflow? Handling a cancel? Cancelling a child? Cancelling an external workflow? Can you point to the TS sample you'd like me to emulate?

hello/hello_activity.py
cretz: Done

hello/hello_activity_choice.py
cretz: Added enums. Also opened https://github.com/temporalio/sdk-python/pull/74.

hello/hello_activity_choice.py
mfateev: I believe in Java we use the target format: 

https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html#forTarget-java.lang.String-

cretz: We do in Go as well. Unfortunately our Rust library does not: https://docs.rs/tonic/latest/tonic/transport/struct.Channel.html. There is some discussion of this at https://github.com/hyperium/tonic/issues/437 and other linked issues.

Basically Hyper requires a URI, even if you override it to be a unix socket. E.g. https://github.com/hyperium/tonic/blob/9ea03c2c44ad6bf76e4141e6fb40870b9b30ab63/examples/src/uds/client.rs#L17-L27

In the future we can offer non-http options and have core parse it. But in the meantime, if we allowed a scheme-less target now, we'd have to have some other way for them to opt in to TLS (like Go and Java require today).

bergundy: > Can you clarify the recommendation? Are you saying in Python we should change to just address or in Core? If the former, are you saying I should prepend the scheme? How do I know whether it should be http or https? I can't use the presence of TLS config because there may be no config needed (unless I made users use an empty one to use TLS which is a bit ugly).

In TS we accept `tls: true | TLSConfig`, you could do the same.

I'd be fine with either that or keeping a URI but validating that the scheme is http or https in core.

No strong preference.

On note for using the HostPort format is that it's consistent with the other SDKs.

hello/hello_activity_choice.py
bergundy: 👍

hello/hello_async_activity_completion.py
cretz: May be subject to a general discussion outside of this PR.

I don't see why we can't ask users to pass all things they need to their activities at class instantiation/registration time (be it a Temporal client, another HTTP client, a DB client, etc).

This would be a new thing for all SDKs which all, today, allow one to do this if they want but don't provide this convenience thing to do it for them because it's not necessary and it's not difficult to do themselves.

hello/hello_child_workflow.py
bergundy: This would be a user defined global.
Something along these lines.


```python
collection = Collection()

@collection.workflow.defn
...

@collection.activity.defn
...

# or

@workflow.defn(collections=[collection])
...

Worker(**collection.worker_config)
```

This roughly follows a common pattern that popular frameworks like celery, fastapi and flask use.

hello/hello_continue_as_new.py
bergundy: I wish we had reviewed this earlier too, seeing these samples now puts this in my face so more suggestions come up.

Nothing critical, if matching asyncio APIs is a goal then I'd say it's fine as is.

hello/hello_continue_as_new.py
bergundy: I like the TS one better, it's much simpler and helps demonstrate that CAN is basically recursion.

Another good sample could be https://docs.temporal.io/typescript/workflows#example, I've recommended this pattern numerous times in the past.

hello/hello_child_workflow.py
cretz: Sounds just as easy to create your own decorator instead of having our decorator do this for you. This is especially true with class and method-based activities that need an instance to be created by the user to be registered anyways.

hello/hello_cron.py
bergundy: 👍

hello/hello_local_activity.py
bergundy: > Are you saying I should mark local activities in Python experimental?

Yes, because of the complexity here and the fact that they're so new even in TS.
The Core part of local activities has barely been used in production making this the least stable feature in the SDK.

> Python is still considered alpha so I wouldn't strictly bother to mention this but since we're pretty confident about the rest of the features I'd consider mentioning.

But also noting this ^

hello/hello_parallel_activity.py
bergundy: I think you'll be surprised by the amount of users that will ask you this question, leaving it up to you to decide

hello/hello_signal.py
bergundy: Yes, right now it's the caller's choice but it's also way too easy to get wrong.

Tomperez98:
Running `hello_activity_choice.py` I get this error message

```
    raise RuntimeError("Failed decoding arguments") from err
RuntimeError: Failed decoding arguments
```

cretz:
@Tomperez98 - Are you using Python 3.10+? We recently fixed some of this in some `main` pushes to the `temporalio` repo, so hopefully it runs w/ latest SDK (this PR will not be merged until after next release since it relies on unreleased features).

Also, we plan to overhaul some of the dataclass type hinting inference, see https://github.com/temporalio/sdk-python/issues/78. That will likely happen before these samples land.

Tomperez98:
> @Tomperez98 - Are you using Python 3.10+? We recently fixed some of this in some `main` pushes to the `temporalio` repo, so hopefully it runs w/ latest SDK (this PR will not be merged until after next release since it relies on unreleased features).
> 
> Also, we plan to overhaul some of the dataclass type hinting inference, see [temporalio/sdk-python#78](https://github.com/temporalio/sdk-python/issues/78). That will likely happen before these samples land.

No. I'm running `python 3.9.12`

hello/hello_activity_choice.py
cretz: PR opened at https://github.com/temporalio/sdk-python/issues/79

hello/hello_continue_as_new.py
cretz: Updated to be the same as the TS one

bergundy (approved):

PR #20, opened 2022-09-27
added hello-activity-test

antmendoza:
## What was changed

- Add test for hello-activity 
- Add conftest.py to be able to inject the client into the tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,6 @@ jobs:
       - run: python -m pip install --upgrade wheel poetry poethepoet
       - run: poetry install
       - run: poe lint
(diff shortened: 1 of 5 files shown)


cretz (commented):
This is great, thanks!

Can you also uncomment the line at the bottom of `.github/workflows/ci.yml` to enable tests in CI? And even add another line after it with the same things but `--workflow-environment time-skipping`?

Also, seems CI is failing due to bad third party type checking. You might need to add `--namespace-packages` to mypy run in pyproject.toml or something. If you can't figure out how to fix the issues that `poe lint-types` is showing, DM me and I can help debug.

cretz (approved):

PR #23, opened 2022-10-26
Sentry interceptor example

nediamond:
## What was changed

added Sentry interceptor example based on discussion in [slack thread](https://temporalio.slack.com/archives/CTT84RS0P/p1666821089048929)

## Why?

Sentry is a popular error monitoring tool & interceptors aren't super intuitive to make from scratch, hopefully this can be a good starting point for folks

## Checklist

1. How was this tested:
Executed locally

2. Any docs updates needed?
not sure

### DIFF
--- a/sentry/interceptor.py
+++ b/sentry/interceptor.py
@@ -1,8 +1,8 @@
 from dataclasses import asdict, is_dataclass
-from typing import Any, Optional, Type
+from typing import Any, Optional, Type, Union
 
 import sentry_sdk
-from temporalio import workflow, activity
+from temporalio import activity, workflow
 from temporalio.worker import (
     ActivityInboundInterceptor,
     ExecuteActivityInput,
@@ -14,7 +14,7 @@ from temporalio.worker import (
 
 
 def _set_common_workflow_tags(
-    info: workflow.Info | activity.Info, scope: sentry_sdk.Scope
+    info: Union[workflow.Info, activity.Info], scope: sentry_sdk.Scope
 ):
     scope.set_tag("temporal.workflow.type", info.workflow_type)
     scope.set_tag("temporal.workflow.id", info.workflow_id)
--- a/sentry/interceptor.py
+++ b/sentry/interceptor.py
@@ -16,10 +16,8 @@ from temporalio.worker import (
 def _set_common_workflow_tags(
     info: workflow.Info | activity.Info,
 ):
-    sentry_sdk.set_tag("temporal.workflow.namespace", info.workflow_namespace)
     sentry_sdk.set_tag("temporal.workflow.type", info.workflow_type)
     sentry_sdk.set_tag("temporal.workflow.id", info.workflow_id)
-    sentry_sdk.set_tag("temporal.workflow.run_id", info.run_id)
 
 
 class _SentryActivityInboundInterceptor(ActivityInboundInterceptor):
@@ -35,6 +33,12 @@ class _SentryActivityInboundInterceptor(ActivityInboundInterceptor):
             sentry_sdk.set_tag("temporal.activity.id", activity_info.activity_id)
             sentry_sdk.set_tag("temporal.activity.type", activity_info.activity_type)
             sentry_sdk.set_tag("temporal.activity.task_queue", activity_info.task_queue)
+            sentry_sdk.set_tag(
+                "temporal.workflow.namespace", activity_info.workflow_namespace
+            )
+            sentry_sdk.set_tag(
+                "temporal.workflow.run_id", activity_info.workflow_run_id
+            )
             try:
                 return await super().execute_activity(input)
             except Exception as e:
@@ -62,6 +66,8 @@ class _SentryWorkflowInterceptor(WorkflowInboundInterceptor):
             workflow_info = workflow.info()
             _set_common_workflow_tags(workflow_info)
             sentry_sdk.set_tag("temporal.workflow.task_queue", workflow_info.task_queue)
+            sentry_sdk.set_tag("temporal.workflow.namespace", workflow_info.namespace)
+            sentry_sdk.set_tag("temporal.workflow.run_id", workflow_info.run_id)
             try:
                 return await super().execute_workflow(input)
             except Exception as e:
--- a/sentry/interceptor.py
+++ b/sentry/interceptor.py
@@ -14,10 +14,10 @@ from temporalio.worker import (
 
 
 def _set_common_workflow_tags(
-    info: workflow.Info | activity.Info,
+    info: workflow.Info | activity.Info, scope: sentry_sdk.Scope
 ):
-    sentry_sdk.set_tag("temporal.workflow.type", info.workflow_type)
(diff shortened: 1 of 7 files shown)


sentry/interceptor.py
```diff
@@ -0,0 +1,75 @@
...
+
+class _SentryActivityInboundInterceptor(ActivityInboundInterceptor):
+    async def execute_activity(self, input: ExecuteActivityInput) -> Any:
+        transaction_name = input.fn.__module__ + "." + input.fn.__name__
+        scope_ctx_manager = sentry_sdk.configure_scope()
+        with scope_ctx_manager as scope, sentry_sdk.start_transaction(
+            name=transaction_name
+        ):
+            sentry_sdk.set_tag("temporal.execution_type", "activity")
+            sentry_sdk.set_tag("temporal.activity.name", input.fn.__name__)
```
cretz, line 24:
```suggestion
            sentry_sdk.set_tag("temporal.activity.name", temporalio.activity.info().activity_type)
```

There are also other tags from info you might want like `activity_id` (so it can be tied w/ the Temporal history/UI), `workflow_type`, `workflow_id`, `workflow_namespace`, `task_queue`, etc.

sentry/interceptor.py
```diff
@@ -0,0 +1,75 @@
...
+    ExecuteWorkflowInput,
+    Interceptor,
+    WorkflowInboundInterceptor,
+    WorkflowInterceptorClassInput,
+)
+
+
+class _SentryActivityInboundInterceptor(ActivityInboundInterceptor):
+    async def execute_activity(self, input: ExecuteActivityInput) -> Any:
+        transaction_name = input.fn.__module__ + "." + input.fn.__name__
```
cretz, line 18:
```suggestion
        transaction_name = input.fn.__module__ + "." + input.fn.__qualname__
```

Sometimes activities are methods of classes. Then again, you could use `temporalio.activity.info().activity_type` which is the registered name with the worker.

sentry/interceptor.py
```diff
@@ -0,0 +1,75 @@
...
+                sentry_sdk.set_context(
+                    "temporal.workflow.info", workflow.info().__dict__
+                )
+                sentry_sdk.capture_exception(e)
+                raise e
+            finally:
+                scope.clear()
+
+
+class SentryInterceptor(Interceptor):
```
cretz, line 63: Can you put a docstring on here describing its purpose?

sentry/interceptor.py
```diff
@@ -0,0 +1,75 @@
...
+
+class _SentryWorkflowInterceptor(WorkflowInboundInterceptor):
+    async def execute_workflow(self, input: ExecuteWorkflowInput) -> Any:
+        transaction_name = input.type.__module__ + "." + input.type.__name__
+        scope_ctx_manager = sentry_sdk.configure_scope()
+        with scope_ctx_manager as scope, sentry_sdk.start_transaction(
+            name=transaction_name
+        ):
+            sentry_sdk.set_tag("temporal.execution_type", "workflow")
+            sentry_sdk.set_tag("temporal.workflow.name", input.type.__name__)
```
cretz, line 46:
```suggestion
            sentry_sdk.set_tag("temporal.workflow.name", temporalio.workflow.info().workflow_type)
```

Same here as activities. There is lots of stuff on `temporalio.workflow.info()` you might want in here

sentry/worker.py
```diff
@@ -0,0 +1,69 @@
...
+        dsn=os.environ.get("SENTRY_DSN"),
+    )
+
+    # Start client
+    client = await Client.connect("localhost:7233")
+
+    # Run a worker for the workflow
+    async with Worker(
+        client,
+        task_queue="hello-activity-task-queue",
```
cretz, line 54:
```suggestion
        task_queue="sentry-task-queue",
```

sentry/worker.py
```diff
@@ -0,0 +1,69 @@
...
+        workflows=[GreetingWorkflow],
+        activities=[compose_greeting],
+        interceptors=[SentryInterceptor()],  # Use SentryInterceptor for error reporting
+    ):
+        result = await client.execute_workflow(
+            GreetingWorkflow.run,
+            "World",
+            id="hello-activity-workflow-id",
+            task_queue="hello-activity-task-queue",
```
cretz, line 63:
```suggestion
            id="sentry-workflow-id",
            task_queue="sentry-task-queue",
```

cretz (commented):
Added a couple of comments. Looks great!

Some other that I think would help:

* Add a README.md in this folder w/ details on how to run (see how OTel sample does it in #18)
* Add a "dependency group" for `sentry` in `pyproject.toml` to add the `sentry_sdk` dependency (see how OTel sample does it in #18)
* Separate the `starter` from the `worker`. We are doing this in all of the non-hello samples, and we do it in all SDKs, to emphasize that workers are often completely separate from starters (see how OTel sample does it in #18)

If the Sentry SDK comes with utilities to capture/assert output in a test, a test in `tests/sentry` would be great. But not required or anything.

nediamond:
will try to implement the tests/readme/worker changes this weekend

cretz:
@nediamond - Wanted to follow up here. I think this is a great sample. Let me know if there's anything I can do to help move it along.

nediamond:
@cretz sorry got busy with life stuff will set some time aside for this before next week - feel clear on what needs to be done thanks

nediamond:
@cretz i reverted the `poetry.lock` file bc it looked like the checks didnt like it - could use your help generating that since it seems like it may be from a difference in my setup or something

added a super basic readme, but i think this is about as much as i can commit to for now

cretz:
> could use your help generating that since it seems like it may be from a difference in my setup or something

> added a super basic readme, but i think this is about as much as i can commit to for now

Ok, I will look to fixing that and improving your README

cretz (approved):

nathanielobrown:
Hey @nediamond and @cretz, I just wanted to share a pitfall I found in this approach for implementing Sentry via interceptors when using threaded activities. Because activities are run in a thread in a `ThreadPoolExecutor` or in a separate process but the interceptor is run in the main thread, the activity code gets a totally different Sentry scope/context than the interceptor. So any context generated _inside the activity_ (such as breadcrumbs, setting tags, context etc.) will **not** be included in the resulting Sentry issue. I also was experiencing some "leakage" of contexts, where transaction names would be wrong (I don't 100% understand what was happening there). I haven't tested with `async` activities, but I feel like _maybe_ they might have the same issue too because Sentry uses `contextvars`?

My solution was to add Sentry handling to a custom decorator that wraps the `@activity.defn` decorator and does Sentry handling in the thread. I left the activity interceptor as a fallback, and set an attribute on the activity function and skip interceptor Sentry handling if this attribute is set.

I just wanted to write this up in case you had any further thoughts on this and as a note for others implementing Sentry with Temporal Python SDK.

cretz:
Hrmm, this is true. I think we need to be putting contextvars from interceptors on to threaded activities. I have opened https://github.com/temporalio/sdk-python/issues/263.

PR #24, opened 2022-10-28
Update default encryption key

lorensr:
Updating default key to match those used in Go and TS.

Using python encryption starter with Go worker works, with: 

- this PR
- this patch:

```
diff --git a/encryption/starter.py b/encryption/starter.py
index 4c39f55..4f26d71 100644
--- a/encryption/starter.py
+++ b/encryption/starter.py
@@ -20,10 +20,10 @@ async def main():
 
     # Run workflow
     result = await client.execute_workflow(
-        GreetingWorkflow.run,
+        "Workflow",
         "Temporal",
         id=f"encryption-workflow-id",
-        task_queue="encryption-task-queue",
+        task_queue="encryption",
     )
     print(f"Workflow result: {result}")
 ```
 
- and commenting out the key id check (Go code appears to use `"test"` for the default key id, but in history it's empty `"encryption-key-id": "",`)

### DIFF
--- a/encryption/codec.py
+++ b/encryption/codec.py
@@ -6,7 +6,7 @@ from cryptography.hazmat.primitives.ciphers.aead import AESGCM
 from temporalio.api.common.v1 import Payload
 from temporalio.converter import PayloadCodec
 
-default_key = base64.b64decode(b"MkUb3RVdHQuOTedqETZW7ra2GkZqpBRmYWRACUospMc=")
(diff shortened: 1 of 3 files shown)


cretz:
@lorensr - You can apply that patch too. Though for the workflow name, just change `@workflow.defn` to `@workflow.defn(name="Workflow")`.

cretz (approved):

PR #30, opened 2022-11-25
add test to hello_cancellation

antmendoza:
## What was changed

Added test to hello_cancellation example

I was wondering if it makes sense to verify that the activity [never_complete_activity](https://github.com/temporalio/samples-python/blob/main/hello/hello_cancellation.py#L20) has been cancelled / has thrown CancelledError, with an interceptor maybe?

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/tests/hello/hello_cancellation_test.py
+++ b/tests/hello/hello_cancellation_test.py
@@ -1,12 +1,9 @@
 import asyncio
 import uuid
 
-from temporalio.api.common.v1 import WorkflowExecution
-from temporalio.api.workflowservice.v1 import (
-    DescribeWorkflowExecutionRequest,
-    DescribeWorkflowExecutionResponse,
-)
+import pytest
 from temporalio.client import Client, WorkflowExecutionStatus, WorkflowFailureError
+from temporalio.exceptions import CancelledError
 from temporalio.worker import Worker
 
 from hello.hello_cancellation import (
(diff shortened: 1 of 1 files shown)


cretz (approved):

tests/hello/hello_cancellation_test.py
```diff
@@ -0,0 +1,39 @@
...
+        activities=[cleanup_activity, never_complete_activity],
+    ):
+        handle = await client.start_workflow(
+            CancellationWorkflow.run,
+            id=(str(uuid.uuid4())),
+            task_queue=task_queue_name,
+        )
+
+        # wait for the activity "never_complete_activity" to heartbeat
+        await asyncio.sleep(2)
```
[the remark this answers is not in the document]
antmendoza: Thanks @cretz, neither I, I didn't know how to wait until the activity heartbeat. 

I think that even if I call describeWorkflow I will have to wait on something / or loop until seing the pending activity. 

How would you do it?

tests/hello/hello_cancellation_test.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+
+        await asyncio.wait_for(
+            wait_for_activity_to_start("never_complete_activity", client, workflow_id),
+            timeout=5,
+        )
+
+        await handle.cancel()
+
+        await asyncio.wait_for(
```
antmendoza, line 42: @cretz I am also waiting until the workflow.status = canceled, let me know if this makes sense.

tests/hello/hello_cancellation_test.py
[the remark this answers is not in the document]
antmendoza: done @cretz , feedback is welcome

tests/hello/hello_cancellation_test.py
cretz: No need to do that, just `await handle.result()` and check the exception like https://github.com/temporalio/sdk-python/blob/e37b50090392fc218bfe051063ab783c051fb0fb/tests/test_client.py#L178-L180

cretz: No need to do that, just `await handle.result()` and check the exception like https://github.com/temporalio/sdk-python/blob/e37b50090392fc218bfe051063ab783c051fb0fb/tests/test_client.py#L178-L180

tests/hello/hello_cancellation_test.py
```diff
@@ -0,0 +1,87 @@
...
+    while not (await has_activity_started(activity_name, workflow_id, client)):
+        await asyncio.sleep(0.2)
+
+
+async def has_activity_started(activity_name, workflow_id, client):
+    response: DescribeWorkflowExecutionResponse = await describe_workflow(
+        client, workflow_id
+    )
+
+    for pending_activity in response.pending_activities:
```
cretz, line 62: I think you can just `handle.describe()` instead of `describe_workflow_execution` and then just use `response.raw_description.pending_activities` here

antmendoza: done

tests/hello/hello_cancellation_test.py
antmendoza: done

cretz (approved):
Great test! Thanks! Merge whenever you'd like.

PR #42, opened 2023-01-10
[BUG] Auto heartbeater death doesn't cancel task

sfgeorge-vl:
## What was changed

Bug fix: Updated auto_heartbeater to cancel the wrapper user Task if `temporalio.activity.heartbeat()` raises `asyncio.CancelledError`.

## Why?

I found through testing that, when the auto-heartbeater received a `asyncio.CancelledError`, it does not Cancel the original wrapper user task.

As noted in the Temporal Python SDK, `temporalio.activity.heartbeat()` receives `asyncio.CancelledError`, and is intended as the mechanism to cancel the running Task.  However in practice, this would only cancel auto_hearbeater's `heartbeat_every()` task without canceling the user's wrapped function/Task.

> Cancellation for asynchronous activities is done via [asyncio.Task.cancel](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel). This means that asyncio.CancelledError will be raised ... [[ 1 ]](https://github.com/temporalio/sdk-python#asynchronous-activities)
> ...
> In order for a non-local activity to be notified of cancellation requests, it must invoke temporalio.activity.heartbeat() [[ 2 ]](https://github.com/temporalio/sdk-python#heartbeating-and-cancellation)

### 1. How was this tested:

See the attached test script [asyncio_exception_test.txt](https://github.com/temporalio/samples-python/files/10387132/asyncio_exception_test.txt), which prints 1 through 10, once a second.  It also occasionally prints “<3” to signify a successful heartbeat.  There is a 20% chance that a heartbeat will instead raise a cancelation, in which case the task of printing 1 through 10 is not interrupted.

**Its expected output:**

```
Start
Ping 0!
Ping 1!
Ping 2!
Ping 3!
<3
Ping 4!
Ping 5!
Ping 6!
Ping 7!
Exception: You are canceled!
End
```

**Its actual output:**

```
Start
Ping 0!
Ping 1!
Ping 2!
Ping 3!
<3
Ping 4!
Ping 5!
Ping 6!
Ping 7!
Exception: You are canceled!
Ping 8!
Ping 9!
End
```

Notice that in the actual behavior, the Task is allowed to continue until completion:

```
Ping 8!
Ping 9!
```

After applying the changes made in this PR, observe the following desired behavior:
1. If the heartbeat_every() asyncio.Task receives a cancel, the wrapped user Task is canceled as well.
2. If the wrapped user Task completes first, the heartbeat_every() asyncio.Task receives a cancel and ends.

### 2. Any docs updates needed?
No need; this is a backwards-compatible change to the auto-heartbeater.

### DIFF
--- a/custom_decorator/activity_utils.py
+++ b/custom_decorator/activity_utils.py
@@ -13,28 +13,26 @@ def auto_heartbeater(fn: F) -> F:
     # available via our wrapper, so we use the functools wraps decorator
     @wraps(fn)
     async def wrapper(*args, **kwargs):
-        done = asyncio.Event()
+        tasks = [asyncio.create_task(fn(*args, **kwargs))]
         # Heartbeat twice as often as the timeout
         heartbeat_timeout = activity.info().heartbeat_timeout
         if heartbeat_timeout:
-            asyncio.create_task(
-                heartbeat_every(heartbeat_timeout.total_seconds() / 2, done)
+            tasks.append(
+                asyncio.create_task(
+                    heartbeat_every(heartbeat_timeout.total_seconds() / 2)
+                )
             )
-        try:
-            return await fn(*args, **kwargs)
-        finally:
-            done.set()
+
+        await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
 
     return cast(F, wrapper)
 
 
 async def heartbeat_every(
-    delay: float, done_event: asyncio.Event, *details: Any
+    delay: float, *details: Any
 ) -> None:
     # Heartbeat every so often while not cancelled
-    while not done_event.is_set():
-        try:
-            await asyncio.wait_for(done_event.wait(), delay)
-        except asyncio.TimeoutError:
-            print(f"Heartbeating at {datetime.now()}")
-            activity.heartbeat(*details)
+    while True:
+        await asyncio.sleep(delay)
+        print(f"Heartbeating at {datetime.now()}")
+        activity.heartbeat(*details)


custom_decorator/activity_utils.py
```diff
@@ -13,28 +13,26 @@ def auto_heartbeater(fn: F) -> F:
...
+                asyncio.create_task(
+                    heartbeat_every(heartbeat_timeout.total_seconds() / 2)
+                )
             )
-        try:
-            return await fn(*args, **kwargs)
-        finally:
-            done.set()
+
+        await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
```
cretz, line 26: Won't the heartbeater run forever?

From the PR description you have:

> If the wrapped user Task completes first, the heartbeat_every() asyncio.Task receives a cancel and ends.

But at https://docs.python.org/3/library/asyncio-task.html#asyncio.wait, it says:

> Unlike [wait_for()](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for), wait() does not cancel the futures when a timeout occurs.

So that means just because one of the tasks here completes, the other doesn't. You can take the pending results and cancel them here though I think and then await them.

cretz (commented):
Thanks for the contribution! It makes sense, I didn't setup the tasks to cascade cancellation to both. Added minor comment and approving to run CI. Will refrain on moving forward until it's marked ready for review.

custom_decorator/activity_utils.py
sfgeorge-vl: Yes! I saw that and that documentation is SUPER confusing. However, it pertains *only* to the `timeout=` kwarg.  When the return_when= condition is satisfied however, the remaining tasks are Task.cancel()-ed (provided that return_when=ALL_COMPLETED or FIRST_EXCEPTION)

```python
coroutine asyncio.wait(aws, *, timeout=None, return_when=ALL_COMPLETED)
```

Here's a convoluted example:

```python
import asyncio


async def fine():
    try:
        await asyncio.sleep(2)
        print('Everything is happy')
    except asyncio.CancelledError as e:
        print('fine() was cancelled.  Everything is /not/ fine.')
        raise e


async def fail(task_to_cancel=None):
    try:
        await asyncio.sleep(1)
    except asyncio.CancelledError:
        print('fail() was cancelled before it could do its thing')
        raise

    if task_to_cancel:
        print('fail() will cancel fine()')
        task_to_cancel.cancel()
    else:
        print('fail() will raise exc')
        raise Exception('Kaboom!')


async def run():
    print('Start')
    fine_task=asyncio.create_task(fine())
    await asyncio.wait(
        [
            fine_task,
            # asyncio.create_task(fail())
            asyncio.create_task(fail(task_to_cancel=fine_task))
        ],
        return_when=asyncio.FIRST_COMPLETED
    )
    print('run() was able to finish without a try/except')
    print('End')

async def main():
    await run()

if __name__ == '__main__':
    asyncio.run(main())
```

The output of `print('run() was able to finish without a try/except')` shows that the caller of asyncio.wait( ..., return_when=asyncio.FIRST_COMPLETED) continues execution in the event of raised cancelations.  It also shows that the caller need not explicitly cleanup / cancel the remaining tasks....asyncio.wait() does this automatically.

```
Start
fail() will cancel fine()
fine() was cancelled.  Everything is /not/ fine.
run() was able to finish without a try/except
End
```

cretz: :+1: If you have confirmed _both_ tasks receive the cancel, that works for me. Feel free to mark this PR ready for review when you feel good about it.

sfgeorge-vl: @cretz Sorry, I may have spoken incorrectly.  There's enough difference between my own tests versus the nice setup in `custom_decorator/` that with the latter, indeed heartbeat_every() does run forever even after `wait_for_cancel_activity()` is canceled.

In short, I'll have more changes & testing to do

sfgeorge-vl:
@cretz apologies for the noise, closing this as a non-bug.  My testing setup was not as end-to-end as yours; in the end, the existing solution works correctly, and my proposed change would have actually _introduced_ the forever-heartbeating concern you raised.

PR #43, opened 2023-01-11
Ensure a strong reference to asyncio Task for auto-heartbeater

sfgeorge-vl:
## What was changed

Assigned the output of `asyncio.create_task()` to a temporary variable.

## Why?

I learned that one should store the return value of any call to `asyncio.create_task()` in order to maintain a strong reference to the Task.

Otherwise the Task can unexpectedly disappear mid-execution, as Python’s garbage collector will occasionally dispose of weak references.

From https://docs.python.org/3/library/asyncio-task.html#creating-tasks

> asyncio.**create_task**(coro, *, name=None, context=None)
> […]
> Important: Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done.

## Checklist

**1. How was this tested:**

- [x] Tested the `custom_decorator/` scenario. ✅ 
```
cd ./custom_decorator
poetry run python worker.py &
poetry run python starter.py
```
- [x] Tested `poetry run python hello/hello_async_activity_completion.py` ✅ 

**2. Any docs updates needed?**

Nope, backwards-compatible change.

### DIFF
--- a/custom_decorator/activity_utils.py
+++ b/custom_decorator/activity_utils.py
@@ -13,28 +13,27 @@ def auto_heartbeater(fn: F) -> F:
     # available via our wrapper, so we use the functools wraps decorator
     @wraps(fn)
     async def wrapper(*args, **kwargs):
-        done = asyncio.Event()
-        # Heartbeat twice as often as the timeout
         heartbeat_timeout = activity.info().heartbeat_timeout
+        heartbeat_task = None
         if heartbeat_timeout:
-            _ = asyncio.create_task(
-                heartbeat_every(heartbeat_timeout.total_seconds() / 2, done)
+            # Heartbeat twice as often as the timeout
+            heartbeat_task = asyncio.create_task(
+                heartbeat_every(heartbeat_timeout.total_seconds() / 2)
             )
         try:
             return await fn(*args, **kwargs)
         finally:
-            done.set()
+            if heartbeat_task:
+                heartbeat_task.cancel()
+                # Wait for heartbeat cancellation to complete
+                await asyncio.wait([heartbeat_task])
 
     return cast(F, wrapper)
 
 
-async def heartbeat_every(
-    delay: float, done_event: asyncio.Event, *details: Any
-) -> None:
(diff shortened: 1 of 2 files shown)


cretz:
> I learned that one should store the return value of any call to asyncio.create_task() in order to maintain a strong reference to the Task.

Ah yes, I knew this, see https://github.com/temporalio/sdk-python/blob/2b4eed8661be533fea21a5b1f721599009283227/temporalio/worker/_workflow_instance.py#L200-L204, I just didn't think about it here.

custom_decorator/activity_utils.py
```diff
@@ -17,7 +17,7 @@ async def wrapper(*args, **kwargs):
         # Heartbeat twice as often as the timeout
         heartbeat_timeout = activity.info().heartbeat_timeout
         if heartbeat_timeout:
-            asyncio.create_task(
+            _ = asyncio.create_task(
```
cretz, line 20: The right answer is technically probably to capture this and cancel-then-await it in a finally instead of using the done event, but this probably works too.

May deserve a comment though on why you added this. Also I assume we can trust the Python interpreter will never get smart enough to know the `_` var can be collected since it is not referenced anymore?

sfgeorge-vl: Great call, the cancel-and-await strategy is not only more stable, but doesn't leave a confusing `_` that would require some explanation.

Do you recommend I similarly adjust my approach for `compose_greeting()`, or just add a comment for that one?

cretz: Can just add comment if you want there (the lifetime of that `_` is probably not long enough for that use case anyways). I think your latest update is plenty elegant.

sfgeorge-vl: Thanks, added.  Please forgive the plagiarism! 😅

cretz (approved):

cretz:
Merged! Thanks!

PR #45, opened 2023-01-14
Features/36 sticky activity queues

### LINKED ISSUE #36: Sample request: Worker-specific task queues
Like https://github.com/temporalio/samples-typescript/tree/main/activities-sticky-queues, aka poor-man's sessions

Oracen:
## What was changed
This PR proposes a Python sample for sticky activity queues, a similar implementation to [the one in the typescript SDK](https://github.com/temporalio/samples-typescript/tree/main/activities-sticky-queues)

## Why?
These changes were made based on the request for a sample demonstrating sticky activity queues

## Checklist

1. Closes #36 

2. How was this tested:
Tested locally against `docker compose` Temporal server

3. Any docs updates needed?
Adding example to project readme


README.md
```diff
@@ -49,6 +49,7 @@ Some examples require extra dependencies. See each sample's directory for specif
     while running.
   * [hello_signal](hello/hello_signal.py) - Send signals to a workflow.
 <!-- Keep this list in alphabetical order -->
+* [activities_sticky_queues](activity_worker) - Uses unique task queues to ensure activities run on specific workers.
```
cretz, line 52:
```suggestion
* [activities_sticky_queues](activities_sticky_queues) - Uses unique task queues to ensure activities run on specific workers.
```

Also, though I know it doesn't match TS examples, I think `activity_sticky_queue` is a better sample/module name here.

activities_sticky_queues/README.md
```diff
@@ -0,0 +1,48 @@
+# Sticky Activity Queues
+
+This sample is a Python implementation of the [Node "Sticky Workers" example](https://github.com/temporalio/samples-typescript/tree/main/activities-sticky-queues), full credit for the design to the authors of that sample. A [sticky execution](https://docs.temporal.io/tasks#sticky-execution) is a job distribution design pattern where all workflow computational tasks are executed on a single worker. In the Golang Temporal SDK this is explicitly supported via the Session option, but in other SDKs a different approach is required.
```
cretz, line 3:
```suggestion
This sample is a Python implementation of the [TypeScript "Sticky Workers" example](https://github.com/temporalio/samples-typescript/tree/main/activities-sticky-queues). A [sticky execution](https://docs.temporal.io/tasks#sticky-execution) is a job distribution design pattern where all workflow computational tasks are executed on a single worker. In the Go and Java SDKs this is explicitly supported via the Session option, but in other SDKs a different approach is required. 
```

activities_sticky_queues/README.md
```diff
@@ -0,0 +1,48 @@
+# Sticky Activity Queues
+
+This sample is a Python implementation of the [Node "Sticky Workers" example](https://github.com/temporalio/samples-typescript/tree/main/activities-sticky-queues), full credit for the design to the authors of that sample. A [sticky execution](https://docs.temporal.io/tasks#sticky-execution) is a job distribution design pattern where all workflow computational tasks are executed on a single worker. In the Golang Temporal SDK this is explicitly supported via the Session option, but in other SDKs a different approach is required. 
+
+Typical use cases for sticky executions include tasks where interaction with a filesystem is required, such as data processing or interacting with legacy access structures. This example will write text files to folders corresponding to each worker, located in
```
cretz, line 5: Missing end of sentence

activities_sticky_queues/README.md
```diff
@@ -0,0 +1,48 @@
+# Sticky Activity Queues
+
+This sample is a Python implementation of the [Node "Sticky Workers" example](https://github.com/temporalio/samples-typescript/tree/main/activities-sticky-queues), full credit for the design to the authors of that sample. A [sticky execution](https://docs.temporal.io/tasks#sticky-execution) is a job distribution design pattern where all workflow computational tasks are executed on a single worker. In the Golang Temporal SDK this is explicitly supported via the Session option, but in other SDKs a different approach is required. 
+
+Typical use cases for sticky executions include tasks where interaction with a filesystem is required, such as data processing or interacting with legacy access structures. This example will write text files to folders corresponding to each worker, located in 
+
+This implementation will follow the Node example:
```
cretz, line 7:
```suggestion
The strategy is:
```

Concerned that the TypeScript example may change out from under us in the future

activities_sticky_queues/README.md
```diff
@@ -0,0 +1,48 @@
...
+- Create a `get_available_task_queue` activity that generates a unique task queue name, `unique_worker_task_queue`.
+- For activities intended to be "sticky", only register them in one Worker, and have that be the only Worker listening on that `unique_worker_task_queue`. This will be run on a series of `FileProcessing` workflows.
+- Execute workflows from the Client like normal. Check the Temporal Web UI to confirm tasks  were staying with their respective worker.
```
cretz, line 10:
```suggestion
- Execute workflows from the Client like normal. Check the Temporal Web UI to confirm tasks were staying with their respective worker.
```

activities_sticky_queues/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+It doesn't matter where the `get_available_task_queue` activity is run, so it can be "non sticky" as per Temporal default behavior. In this demo, `unique_worker_task_queue` is simply a `uuid` initialized in the Worker, but you can inject smart logic here to uniquely identify the Worker, [as Netflix did](https://community.temporal.io/t/using-dynamic-task-queues-for-traffic-routing/3045). Our example differs from the Node sample by running across 5 unique task queues.
+
+Activities have been artificially slowed with `time.sleep(3)` to simulate slow activities.
+
+### Running This Sample
+
+1. `temporal server start-dev` to start [Temporal Server](https://github.com/temporalio/cli/#installation).
```
cretz, line 18: I think we should follow the other sample READMEs where we just point to the primary README for prerequisites and just have the starting of worker and starter here.

activities_sticky_queues/starter.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+async def main():
+    # Connect client
+    client = await Client.connect("localhost:7233")
+
+    # Start 10 concurrent workflows
+    futures = []
+    for _ in range(10):
+        handle = client.execute_workflow(
```
cretz, line 16:
```suggestion
        coro = client.execute_workflow(
```

Or "result" or something. Just to show it's not a `handle` like `start_workflow` would return.

activities_sticky_queues/starter.py
```diff
@@ -0,0 +1,27 @@
...
+            FileProcessing.run, id=str(uuid4()), task_queue="distribution-queue"
+        )
```
cretz, line 18:
```suggestion
    for _ in range(10):
        handle = client.execute_workflow(
            FileProcessing.run,
            id=f"activity_sticky_queue-workflow-id-{i}",
            task_queue="activity_sticky_queue-distribution-queue",
        )
```

Fixed workflow IDs and qualify the task queue with the sample name (matches our other samples)

activities_sticky_queues/tasks.py
```diff
@@ -0,0 +1,111 @@
+import asyncio
+import random
+from dataclasses import dataclass
+from datetime import timedelta
+from hashlib import md5
```
cretz, line 5: Would prefer sha256

activities_sticky_queues/tasks.py
```diff
@@ -0,0 +1,111 @@
...
+    path: str
+
+
+async def _get_available_task_queue(available_queues: List[str]) -> str:
+    """Randomly assign the job to a queue"""
+    return random.choice(available_queues)
+
+
+@activity.defn
+async def download_file_to_worker_filesystem(details: DownloadObj):
```
cretz, line 27:
```suggestion
async def download_file_to_worker_filesystem(details: DownloadObj) -> None:
```

Would be considered a bit pedantic for most Python functions, but explicit return type is a good practice on activities for caller and return value serialization reasons.

activities_sticky_queues/tasks.py
```diff
@@ -0,0 +1,111 @@
...
+    """Simulates downloading a file to a local filesystem"""
+    activity.logger.info(f"Downloading ${details.url} and saving to ${details.path}")
+    # Here is where the real download code goes
```
cretz, line 30:
```suggestion
    # Here is where the real download code goes. Developers should be careful
    # not to block an async activity. If there are concerns about blocking download
    # or disk IO, developers should use loop.run_in_executor or change this activity
    # to be synchronous. Also like for all non-immediate activities, be sure to
    # heartbeat during download.
```

activities_sticky_queues/worker.py
```diff
@@ -0,0 +1,66 @@
...
+async def main():
+    # Uncomment the line below to see logging
+    # logging.basicConfig(level=logging.INFO)
+
+    # Comment line to see non-deterministic functionality
+    random.seed(667)
+
+    # Create random task queues and create parameterised workflow
+    task_queues: List[str] = [str(UUID(int=random.getrandbits(128))) for _ in range(5)]
+    get_available_task_queue = tasks.build_nonsticky_activity(task_queues)
```
cretz, line 24:
```suggestion
    @activity.defn
    async def get_available_task_queue() -> str:
```

Might as well just put this right in here. Then also add that stub to the `tasks.py` file like:

```python
    @activity.defn
    async def get_available_task_queue() -> str:
        """Just a stub for typedworkflow invocation."""
        raise NotImplementedError
```

activities_sticky_queues/worker.py
```diff
@@ -0,0 +1,66 @@
...
+
+async def main():
+    # Uncomment the line below to see logging
+    # logging.basicConfig(level=logging.INFO)
+
+    # Comment line to see non-deterministic functionality
+    random.seed(667)
+
+    # Create random task queues and create parameterised workflow
+    task_queues: List[str] = [str(UUID(int=random.getrandbits(128))) for _ in range(5)]
```
cretz, line 23:
```suggestion
    task_queues = [f"activity_sticky_queue-host-{UUID(int=random.getrandbits(128))}" for _ in range(5)]
```

Just to keep that prefix

activities_sticky_queues/tasks.py
```diff
@@ -0,0 +1,111 @@
...
+        workflow.logger.info("Searching for available worker")
+        unique_worker_task_queue = await workflow.execute_activity(
+            activity="get_available_task_queue",
+            start_to_close_timeout=timedelta(seconds=10),
+        )
+        workflow.logger.info(f"Matching workflow to worker {unique_worker_task_queue}")
+
+        filename = str(workflow.uuid4())
+        download_path = str(LOCAL_PATH / unique_worker_task_queue / filename)
+        download_params = DownloadObj(url="http://temporal.io", path=download_path)
```
cretz, line 88: Can be a bit confusing to the reader for the workflow to manage file paths for the activities. I wonder if the first activity wants to return the downloaded path for use by the other two?

cretz (commented):
Thanks for this! Lots of users have been asking and I'll be happy to not have to keep pointing to the TypeScript one. Added a few comments.

Also, not required, but if you want you can add tests.

Oracen:
Will integrate this feedback as soon as I get a chance

Oracen:
R.E. tests I felt it was a bit overkill for slapping together a small example, but if I don't have them done before tonight I'll have them in tomorrow. Mostly I noticed nobody else had tests, and I didn't want to be a tryhard :laugh:

tests/activities_sticky_queues/activities_sticky_queues_activity_test.py
```diff
@@ -0,0 +1,41 @@
+import uuid
```
cretz, line 1: You don't have to test these FS functions if you want, what you had before is just fine. The only tests I was thinking was to test the workflow behavior (can even mock out the activities w/ different definitions here).

Oracen: I might hang on to them seeing as they're already written, I was using them to help me check some refactoring decisions anyway

cretz (approved):
Looks great, will merge if/when CI passes

tests/activity_sticky_queues/activity_sticky_worker_workflow_test.py
```diff
@@ -0,0 +1,167 @@
...
+    # p(pass| should fail) = 0.333 ^ (n_workers * n_iters)~ 6e-8
+    workflow_ids = [str(uuid.uuid4()) for _ in range(5)]
+    async with worker_distribution, worker_1, worker_2, worker_3:
+        for workflow_id in workflow_ids:
+            result = await client.execute_workflow(
+                tasks.FileProcessing.run,
+                id=workflow_id,
+                task_queue=task_queue_name_distribution,
+                retry_policy=retry_policy,
+                run_timeout=run_timeout,
```
cretz, line 155: Should not set these values unless you really need them

cretz (commented):
Looks like CI is failing, added comment where you don't need to set retry policy or timeout

tests/activity_sticky_queues/activity_sticky_worker_workflow_test.py
Oracen: These were included due to `should fail/does fail` leaving the test server hanging - recommended implementation to keep it moving?

Oracen: These were included due to `should fail/does fail` leaving `poe test` hanging. The test cases pass (with warnings) but the test server never cleans up. What's the recommended setup to prevent the test server from choking on deliberately failed test cases? Pushing up removal now in case you need a local replication

cretz: :+1: Will investigate

cretz: Just ran the tests, and it seems it's because the `get_available_task_queue` activity retries forever with a `NotImplementedError`. By default activities will retry forever, and that can halt a worker shutdown. At a quick glance it seems `test_workflow_fails_without_task_queue_fn` uses the not-implemented `get_available_task_queue` which retries forever on its error. I can investigate a bit deeper later if that's not clear.

Oracen:
Is there a way we can flag a NotImplemented error as an ActivityError? The
thought here is to make sure people porting their code across to temporal
don't need to replace all their exceptions and exception handlers. I did
see some argument for marking exceptions as non returnable but it didn't
seem to change the outcome of the test

cretz:
> Is there a way we can flag a NotImplemented error as an ActivityError?

All exceptions thrown from activities are activity errors to the workflow. But if you have a retry policy that says to retry activity errors forever then the workflow never sees it. Or are you asking if we can flag a `NotImplementedError` as a _non-retryable_ activity error? You should be able to just add `"NotImplementedError"` to the set `non_retryable_error_types` of the retry policy when invoking the activity, but that is per call of course. I don't think we want to globally assume on the user's' behalf that any Python errors are implicitly non-retryable (we don't in other SDKs).

cretz (approved):
Looks good! I hope you don't mind if I merge main into your branch

tests/activity_sticky_queues/activity_sticky_worker_workflow_test.py
```diff
@@ -0,0 +1,135 @@
...
+                tasks.FileProcessing.run,
+                id=str(uuid.uuid4()),
+                task_queue=task_queue_name,
+                retry_policy=RetryPolicy(
+                    maximum_attempts=1,
+                    maximum_interval=timedelta(milliseconds=50),
+                ),
+                run_timeout=timedelta(milliseconds=50),
+            )
+        mock_del.assert_called_once()
```
cretz, line 82: Looks like CI is failing because `delete_file` isn't call for whatever reason

Oracen: This test is passing locally, I'll have a look at the setup of the CI and see if I can figure out what might be missing. I'll also rebuild the env from scratch

Oracen: Merged the latest changes and got a local replication

Oracen:
Moved the retry spec into the worker so the activity would fail once and proceed to the `finally` block. Latest updates to the branch pulled in, tests passing locally

tests/activity_sticky_queues/activity_sticky_worker_workflow_test.py
```diff
@@ -0,0 +1,127 @@
...
+                id=workflow_id,
+                task_queue=task_queue_name_distribution,
+            )
+            assert result == CHECKSUM
+    # Check all events take place on the same random worker
+    workflow_executions = {
+        check_sticky_activity_count(
+            item.to_json_dict()["events"], task_queue_name_distribution
+        )
+        async for item in client.list_workflows().map_histories()
```
cretz, line 123: Our unit tests run on the regular Temporal server and _also_ run with the time-skipping test server which does not implement list workflows (sorry!). You can change the top to:

```python
async def test_worker_runs(client: Client):
    if env.supports_time_skipping:
        pytest.skip("Test server doesn't support listing workflows")
```

And then don't forget the formatter. We're almost there, sorry for all of the CI checks :-)

Oracen: Nothing to apologise for! Better it breaks in CI than PROD

tests/activity_sticky_queues/activity_sticky_queues_activity_test.py
```diff
@@ -0,0 +1,35 @@
...
+
+RETURNED_PATH = "valid/path"
+tasks._get_delay_secs = mock.MagicMock(return_value=0.0001)
+tasks._get_local_path = mock.MagicMock(return_value=Path(RETURNED_PATH))
+
+
+async def test_download_activity():
+    worker_id = "an-id"
+    workflow_uuid = "uuid"
+    want = "/".join([RETURNED_PATH, worker_id, workflow_uuid])
```
cretz, line 14: This is failing on Windows due to dir separator differences :-) Can use `pathlib` or something

Oracen: Done

Oracen: Done

Oracen:
Part of the fix didn't push, checked the fail reason and it should be corrected. Let me know if we need to resolve the path to handle OS paths as I can't test against Windows locally

cretz:
Everything looks great! Merging now. Thanks again, this will help a lot of people.

PR #46, opened 2023-01-20
Improve sentry workflow & activity interceptors

nediamond:
Workflow interceptor was causing sandbox restriction warnings: `__call__ on threading.Lock restricted` & current sentry sdk usage seemed like it might have been causing threading issues with activities

## What was changed
Removed sentry workflow interceptor & improved the wrapper logic to more closely align with their documentation.

## Why?
Current implementation does not work as intended in production.

## Checklist

1. How was this tested:
Currently being used in our remote environments

### DIFF
--- a/sentry/interceptor.py
+++ b/sentry/interceptor.py
@@ -1,77 +1,38 @@
 from dataclasses import asdict, is_dataclass
-from typing import Any, Optional, Type, Union
+from typing import Any
 
-import sentry_sdk
-from temporalio import activity, workflow
+from sentry_sdk import Hub, capture_exception, set_context, set_tag
+from temporalio import activity
 from temporalio.worker import (
     ActivityInboundInterceptor,
     ExecuteActivityInput,
-    ExecuteWorkflowInput,
     Interceptor,
-    WorkflowInboundInterceptor,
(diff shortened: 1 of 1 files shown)


cretz:
I think you should leave the workflow interceptor, make sure it doesn't report failures when `workflow.unsafe.is_replaying()` and add `with workflow.unsafe.imports_passed_through():` around the sentry import.

sentry/interceptor.py
```diff
@@ -12,66 +11,60 @@
...
+with workflow.unsafe.imports_passed_through():
+    from sentry_sdk import Hub, capture_exception, set_context, set_tag
 
-def _set_common_workflow_tags(
-    info: Union[workflow.Info, activity.Info], scope: sentry_sdk.Scope
-):
-    scope.set_tag("temporal.workflow.type", info.workflow_type)
-    scope.set_tag("temporal.workflow.id", info.workflow_id)
+
+def _set_common_workflow_tags(info: workflow.Info | activity.Info):
```
cretz, line 18: We'll require the explicit `Union` because we work on older Python versions

cretz (approved):
Looks good!

PR #47, opened 2023-01-24
unrestrict sentry workflow interceptor sdk call

nediamond:
even with suggested changes from https://github.com/temporalio/samples-python/pull/46 we are still getting `__call__ on threading.Lock restricted` warnings - this removes workflow interceptor to prevent those & related issues


cretz:
Mentioned in Slack, but Sentry is making non-deterministic calls in the sandbox by using thread locks. You can surround the non-deterministic calls with `with temporalio.workflow.unsafe.sandbox_unrestricted():`.

cretz (approved):

PR #51, opened 2023-02-13
Adds a Cancel Activity example

rachfop:
## What was changed
Adds an example on how to Cancel an Activity with the [asyncio task cancel](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel).

## Why?
For documentation purposes.
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

Will snipsync into documenation.


cretz:
Why closed? (and if not reopening, can we delete the branch...may want to do other branches on a fork instead of in this repo)

PR #39, opened 2023-01-04
Adds Flask and Fast APIs

rachfop:
## What was changed
Adds Flask and FastAPI samples to the samples repo.
## Why?

Shows how to create durable workflows with Flask and FastAPI.

## Checklist

1. Closes 

2. How was this tested:

```
poe [test, lint, lint-types, format]
black .
isort .
```

3. Any docs updates needed?

Added readme's in each directory.
Updated main readme with reference.

PR #52, opened 2023-02-15
Adds Starlette

rachfop:
## What was changed
Adds the Starlette Framework as a code sample.

## Why?

Users were asking for a Starlette example.

## Checklist

`poe format`
adds a tool dependcy
adds a readme
adds a test

2. How was this tested:

In the `tests/starlette` folder

3. Any docs updates needed?

      
     Updated README.


starlette/README.md
```diff
@@ -0,0 +1,42 @@
+# Starlette Temporal Workflow Template
+
+This repository provides settings for [Starelette](https://www.starlette.io/) and the Temporal Python SDK for durable workflows.
```
cretz, line 3:
```suggestion
This repository provides settings for [Starlette](https://www.starlette.io/) and the Temporal Python SDK for durable workflows.
```

starlette/starter.py
```diff
@@ -0,0 +1,23 @@
...
+from temporalio.client import Client
+from worker import SayHello
+
+from starlette.applications import Starlette
+from starlette.responses import JSONResponse
+from starlette.routing import Route
+
+
+async def say_hello_endpoint(request):
+    client = await Client.connect("localhost:7233")
```
cretz, line 12: Similar to the other PR at #39, we need to not connect a client on every endpoint but rather reuse one. This needs to use the accepted way to add/inject stateful dependencies into request handlers. https://www.starlette.io/applications/#storing-state-on-the-app-instance may help.

starlette/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Expected output:
+
+```bash
+{"response":"Hello, World!"}
+```
+
+## Further Reading
+
+- [Starelette documentation](https://www.starlette.io/)
```
cretz, line 41:
```suggestion
- [Starlette documentation](https://www.starlette.io/)
```

tests/starlette/hello_activity_test.py
```diff
@@ -0,0 +1,45 @@
+import uuid
```
cretz, line 1: By the way I don't mind, since this is just a hello workflow from somewhere else, that you just get rid of these copy/pasted tests. Having said that, showing how to use Starlette-based tests would be nice.

starlette/README.md
```diff
@@ -0,0 +1,42 @@
+# Starlette Temporal Workflow Template
+
+This repository provides settings for [Starlette](https://www.starlette.io/) and the Temporal Python SDK for durable workflows.
```
cretz, line 3: Question now that tests are gone and all this does is set a client and then call Temporal on route - is there any value in the sample? Can't anyone with any framework create a Temporal client and then use it in a handler? I don't mind adding it, but it seems unnecessary to build this same thing for every framework that does the same thing.

cretz:
@rachfop - Why closed/deleted? Samples didn't seem worth it or was there another reason?

PR #57, opened 2023-02-27
Added sample for demonstrating how to patch a workflow

ktenzer:
## What was changed

Added a sample for how to use patch to version workflows

## Why?

Customer reached out and asked for an example. We didn't have any so I created one.

## Checklist

1. Closes 

2. How was this tested:

Ran using poetry against temporalite and linted using poe

3. Any docs updates needed?

No


hello/hello_patched.py
```diff
@@ -0,0 +1,113 @@
...
+        greeting = await workflow.execute_activity(
+            compose_greeting,
+            ComposeGreetingInput("Hello", name),
+            start_to_close_timeout=timedelta(seconds=60),
+        )
+        await asyncio.sleep(120)
+        return greeting
+
+
+#    async def run(self, name: str) -> str:
```
cretz, line 50: Other examples like https://github.com/temporalio/samples-typescript/tree/main/patching-api have all of the workflows as different classes and just swap out as needed. We did this in Python tests too: https://github.com/temporalio/sdk-python/blob/cd8cbf95a898b25ff3f3c34fe78ddacdd65855a1/tests/worker/test_workflow.py#L2003-L2092. That way you don't have to have commented out code, you can just let maybe a CLI arg choose which one is registered.

hello/hello_patched.py
```diff
@@ -0,0 +1,113 @@
...
+        except exceptions.WorkflowAlreadyStartedError:
+            print(f"Workflow already running")
+            workflow_result = await client.get_workflow_handle(
+                "hello-patched-workflow-id"
+            ).result()
+            print(f"Successful workflow result: {workflow_result}")
+
+    # while(True):
+    #     await asyncio.sleep(30)
+    #     print(f"Working waiting on tasks...")
```
cretz, line 109: Why this code?

hello/hello_patched.py
```diff
@@ -0,0 +1,113 @@
+import asyncio
```
cretz, line 1: Need a README entry for this file (`/README.md` and `/hello/README.md`)

hello/hello_patched.py
ktenzer: that was commented out stuff, I will remove it...wasnt used

hello/hello_patched.py
ktenzer: Ok i broke out the files into a patch directory w/readme

hello/patch/README.md
```diff
@@ -0,0 +1,9 @@
+# Patch Workflow Sample
```
cretz, line 1: Sorry to keep being pedantic, but multi-file examples aren't really good for "hello". Can you make this a top-level `patching` example with starter, worker, and these files and what not? Otherwise, come back to single file.

hello/patch/README.md
```diff
@@ -0,0 +1,9 @@
...
+This sample will demonstrate how to version a workflow. The sample workflow will execute a activity and fire a timer.
```
cretz, line 2: As all of the other READMEs do, make sure to point people to the primary README for prerequisites

hello/patch/README.md
```diff
@@ -0,0 +1,9 @@
+# Patch Workflow Sample
+This sample will demonstrate how to version a workflow. The sample workflow will execute a activity and fire a timer. 
+
+- First run the v1 workflow
```
cretz, line 4: You'll have to give people commands to run things

hello/patch/README.md
ktenzer: re-implemented it with args to run either v1 or v2 of the workflow

hello/patch/README.md
ktenzer: now its straight-forward with cmd args

hello/patch/README.md
ktenzer: N/A

hello/hello_patch.py
```diff
@@ -0,0 +1,149 @@
...
+        return greeting
+
+
+# V2 of patch-workflow using patched where we have changed newly started
+# workflow behavior without changing the behavior of currently running workflows
+@workflow.defn(name="patch-workflow")
+class MyWorkflowPatched:
+    @workflow.run
+    async def run(self, name: str) -> str:
+        print(f"Running patch-workflow with parameter %s" % name)
```
cretz, line 49: Should not use print from inside a workflow. Can use workflow logger if you want.

cretz (commented):
This looks good enough for me. Technically a better demonstration workflow would be like https://github.com/temporalio/samples-typescript/tree/main/patching-api/src has where it adds sleeps for days to demonstrate old workflows can still run with new code, but this is fine.

Don't forget to add README list entries to `/README.md` and `/hello/README.md` for this example

hello/hello_patch.py
ktenzer: ok ill swap that

cretz (approved):

Issue #61, opened 2023-03-16
Sample request: Schedules


cretz:
### Describe the solution you'd like

Need sample showing schedules

rachfop:
I beleive this can be closed.

cretz:
Thanks!

PR #62, opened 2023-03-23
Adds Schedules Features

rachfop:
## What was changed
Updated Temporal Repo.
Added Schedules. 

## Why?

It was requested:

https://github.com/temporalio/samples-python/issues/61

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


schedules/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+Start the Schedules file, then run a feature.
+
+```bash
+poetry run python schedules/start_schedules.py
+poetry run python schedules/backfill_schedule.py
+poetry run python schedules/delete_schedule.py
+poetry run python schedules/describe_schedule.py
+poetry run python schedules/toggle_schedule.py
+poetry run python schedules/update_schedule.py
```
cretz, line 28: So many examples, awesome

schedules/README.md
cretz, line 1: Like other examples here, can you both add a part here saying to look at the main README for prerequisites (e.g. server setup) and add an entry in the main README pointing to here?

poetry.lock
cretz, line 1: We shouldn't change this file unless we're updating `pyproject.toml`. But I do think we need to update `temporalio` to 1.1.0 in `pyproject.toml`

So, revert this file. Then change `temporalio` to be `1.1.0`. Then do a `poetry update temporalio` to update that specific dependency.

schedules/backfill_schedule.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+async def main():
+    client = await Client.connect("localhost:7233")
+    handle = client.get_schedule_handle(
+        "workflow-schedule-id",
+    )
+    now = datetime.utcnow()
+    if now.second == 0:
+        now += timedelta(seconds=1)
```
cretz, line 14: Why make this change? We should be wary of blindly copying test code as sample code. We need to make sure we understand every line we are writing as samples.

schedules/backfill_schedule.py
```diff
@@ -0,0 +1,27 @@
...
+    await handle.backfill(
+        ScheduleBackfill(
+            start_at=now,
+            end_at=now,
+            overlap=ScheduleOverlapPolicy.ALLOW_ALL,
+        ),
+    )
+
+    print(f"Result: {handle}")
```
cretz, line 23: Why print out the handle from get handle?

schedules/delete_schedule.py
```diff
@@ -0,0 +1,19 @@
...
+from temporalio.client import Client
+
+
+async def main():
+    client = await Client.connect("localhost:7233")
+    handle = client.get_schedule_handle(
+        "workflow-schedule-id",
+    )
+    desc_handle = await handle.describe()
+    print(f"{desc_handle.schedule}")
```
cretz, line 12: Why print out the schedule? If we do want to, might as well add a prefix word or something instead of just random dump of data to screen.

schedules/backfill_schedule.py
```diff
@@ -0,0 +1,27 @@
...
+
+async def main():
+    client = await Client.connect("localhost:7233")
+    handle = client.get_schedule_handle(
+        "workflow-schedule-id",
+    )
+    now = datetime.utcnow()
+    if now.second == 0:
+        now += timedelta(seconds=1)
+    await handle.backfill(
```
cretz, line 15: Backfilling a schedule for just "now" makes little sense. You want to backfill in the past. And usually over a range.

schedules/delete_schedule.py
```diff
@@ -0,0 +1,19 @@
...
+
+    await handle.delete()
+    print(f"State: {desc_handle.schedule.state}")
```
cretz, line 15: A handles state is not automatically updated after delete, you would have to refetch it. But then you would get a not-found. Regardless, doesn't have value to print the pre-deleted state of a schedule after you've deleted it

schedules/describe_schedule.py
```diff
@@ -0,0 +1,18 @@
...
+
+async def main():
+    client = await Client.connect("localhost:7233")
+    handle = client.get_schedule_handle(
+        "workflow-schedule-id",
+    )
+
+    desc = await handle.describe()
+
+    print(f"Returns the memo: {desc.schedule.state.note}")
```
cretz, line 14: Note is not "memo" (we use that word for other purposes)

schedules/run_worker.py
```diff
@@ -0,0 +1,21 @@
...
+from temporalio.worker import Worker
+from your_activities import your_activity
+from your_workflows import YourSchedulesWorkflow
+
+
+async def main():
+    client = await Client.connect("localhost:7233")
+    worker = Worker(
+        client,
+        task_queue="my-task-queue",
```
cretz, line 13: Like all other samples, can you make the task queue name unique to the sample?

schedules/start_schedules.py
```diff
@@ -0,0 +1,58 @@
...
+                # intervals=[
+                #    ScheduleIntervalSpec(
+                #        every=timedelta(days=10),
+                #        offset=timedelta(days=2),
+                #    )
+                # ],
+                # cron_expressions=["0 12 * * MON"],
+                # skip=[ScheduleCalendarSpec(year=(ScheduleRange(2050),))],
+                # start_at=datetime(2060, 7, 8, 9, 10, 11, tzinfo=timezone.utc),
+                # jitter=timedelta(seconds=80),
```
cretz, line 48: No need for commented out code in samples

schedules/start_schedules.py
cretz, line 1: This is the only filename using `_schedules` plural, but it's only starting a single schedule

schedules/start_schedules.py
```diff
@@ -0,0 +1,58 @@
...
+            ),
+            state=ScheduleState(
+                note="Here's a note on my Scheduled Workflows", paused=False
```
cretz, line 51:
```suggestion
                note="Here's a note on my Schedule"
```

schedules/toggle_schedule.py
```diff
@@ -0,0 +1,20 @@
...
+    client = await Client.connect("localhost:7233")
+    handle = client.get_schedule_handle(
+        "workflow-schedule-id",
+    )
+
+    await handle.pause(note="Pausing the schedule for now")
+
+    desc = await handle.describe()
+
+    print(f"Describe the schedule's state: {desc.schedule.state}")
```
cretz, line 16: "Describe the schedule's state:" is more like a command asking someone to do something rather than a description of what you're dumping.

Also, is it worth showing the entire state on these samples?

schedules/update_schedule.py
```diff
@@ -0,0 +1,32 @@
...
+        input: ScheduleUpdateInput, timeout_minutes: int = 7
+    ) -> ScheduleUpdate:
+        schedule_action = input.description.schedule.action
+        schedule_action.task_timeout = timedelta(minutes=timeout_minutes)
+        schedule_action.args = ["my new schedule arg"]
+
+        return ScheduleUpdate(schedule=input.description.schedule)
+
+    await handle.update(update_schedule_simple)
+    await handle.trigger()
```
cretz, line 23: Why trigger on update sample? You don't trigger on create.

schedules/update_schedule.py
```diff
@@ -0,0 +1,32 @@
...
+    client = await Client.connect("localhost:7233")
+    handle = client.get_schedule_handle(
+        "workflow-schedule-id",
+    )
+
+    async def update_schedule_simple(
+        input: ScheduleUpdateInput, timeout_minutes: int = 7
+    ) -> ScheduleUpdate:
+        schedule_action = input.description.schedule.action
+        schedule_action.task_timeout = timedelta(minutes=timeout_minutes)
```
cretz, line 17: We should never really show task timeout update in a sample. We discourage altering this value.

schedules/update_schedule.py
```diff
@@ -0,0 +1,32 @@
...
+
+        return ScheduleUpdate(schedule=input.description.schedule)
+
+    await handle.update(update_schedule_simple)
+    await handle.trigger()
+
+    async def schedule_count() -> int:
+        return len([i async for i in await client.list_schedules()])
+
+    print(f"Schedule count: {await schedule_count()}")
```
cretz, line 28: Why dump schedule count? You didn't affect schedule count by updating it.

schedules/update_schedule.py
cretz: Yes, you want to make sure that all of the lints pass.

In this case, the `action` is of that type from MyPy's POV because it's the base class `ScheduleAction`. Right now we only have `ScheduleActionStartWorkflow` but we'll have more in the future. You'll need to use something like `if isinstance(schedule_action, ScheduleActionStartWorkflow):`

cretz:
CI is failing due to a Poetry bug, I will try to upgrade temporalio and fix it in another PR.

schedules/toggle_schedule.py
cretz: This print wording still exists on `trigger_schedule.py`

schedules/delete_schedule.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+async def main():
+    client = await Client.connect("localhost:7233")
+    handle = client.get_schedule_handle(
+        "workflow-schedule-id",
+    )
+
+    await handle.delete()
+    print(f"Schedule deleted.")
```
cretz, line 13: Sometimes you put punctuation on your print statements, sometimes you don't

schedules/start_schedule.py
```diff
@@ -0,0 +1,58 @@
...
+    ScheduleCalendarSpec,
+    ScheduleIntervalSpec,
+    ScheduleRange,
+    ScheduleSpec,
+    ScheduleState,
+)
+from your_workflows import YourSchedulesWorkflow
+
+
+async def main():
```
cretz, line 17: You're inconsistent with whether you print something (e.g. delete or pause) or not (e.g. create or update)

schedules/your_workflows.py
```diff
@@ -0,0 +1,18 @@
+from datetime import timedelta
+
+from temporalio import workflow
+from your_dataobject import YourParams
```
cretz, line 4: I think this can fit under `your_activities` since it's only activity params. Even if you do keep it in its own files, might as well pass it through also.

schedules/list_schedule.py
```diff
@@ -0,0 +1,19 @@
...
+async def main() -> None:
+    client = await Client.connect("localhost:7233")
+
+    async def list_schedule_simple() -> None:
+        async for schedule in await client.list_schedules():
+            schedule_info = schedule.info
+            if isinstance(schedule_info, ScheduleListInfo):
+                print(f"List Schedule Info: {schedule_info}.")
+
+    return await list_schedule_simple()
```
cretz, line 15: Why use a separate local function instead of just calling `async for` right here?

schedules/list_schedule.py
```diff
@@ -0,0 +1,19 @@
...
+from temporalio.client import Client, ScheduleListInfo
+
+
+async def main() -> None:
+    client = await Client.connect("localhost:7233")
+
+    async def list_schedule_simple() -> None:
+        async for schedule in await client.list_schedules():
+            schedule_info = schedule.info
+            if isinstance(schedule_info, ScheduleListInfo):
```
cretz, line 12: Why check `isinstance` here?

schedules/README.md
```diff
@@ -0,0 +1,27 @@
+# Schedules Samples
+
+These samples show how to schedule a Workflow Execution and control certain action.
+
+To run, first see [README.md](../README.md) for prerequisites. Then, run the following from this directory to run the
+`hello_activity.py` sample:
```
cretz, line 6: > `hello_activity.py` sample

Copy/pasta

schedules/README.md
```diff
@@ -0,0 +1,27 @@
...
+Replace `start_schedule.py` in the command with any other example filename to run it instead.
+
+    poetry run python backfill_schedule.py
+    poetry run python delete_schedule.py
+    poetry run python describe_schedule.py
+    poetry run python list_schedule.py
+    poetry run python pause_schedule.py
+    poetry run python update_schedule.py
+
+- [x] create: Creates a new Schedule. Newly created Schedules return a Schedule ID to be used in other Schedule commands.
```
cretz, line 20: Why checked checkboxes instead of just bullets?

schedules/README.md
```diff
@@ -0,0 +1,27 @@
...
+- [x] backfill: Executes Actions ahead of their specified time range.
```
cretz, line 21: Not "ahead" per se.

cretz (commented):
Looks good, a couple of minor comments

cretz (approved):
Looks good, thanks!

PR #63, opened 2023-03-24
Fixes Task Queue name on the Encryption

rachfop:
## What was changed
Matched the Task Queue name on the starter to run with the Worker.

## Why?
No Workers will run until the TQ name matches.

## Checklist

1. Closes 

2. How was this tested:
Ran locally. 

3. Any docs updates needed?

### DIFF
--- a/encryption/starter.py
+++ b/encryption/starter.py
@@ -23,7 +23,7 @@ async def main():
         GreetingWorkflow.run,
         "Temporal",
         id=f"encryption-workflow-id",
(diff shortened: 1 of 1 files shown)


cretz (approved):
Change looks good.

Ug, so Poetry 1.4.1 is broken causing CI breaks. We need to find a way to install specific Poetry version 1.4.0 or wait on https://github.com/python-poetry/poetry/pull/7694. I will look into this soon hopefully.

PR #65, opened 2023-03-27
3 more Python samples

### LINKED ISSUE #53: Sample request: Activities as instance methods
### Describe the solution you'd like

We kinda have this with `/hello/hello_async_activity_completion.py` but we need something clearer that shows usage of something like a DB client

### LINKED ISSUE #58: Sample request: Prometheus
### Describe the solution you'd like

Simple sample showing the configuration of Prometheus metrics

### LINKED ISSUE #59: Sample request: Patching
### Describe the solution you'd like

Since it's not documented yet, we need to at least have a sample of patching

cretz:
## What was changed

* Sample for how to use activity instance methods (closes #53)
* Sample for configuring Prometheus (closes #58)
* Sample for patching (closes #59)

### DIFF
--- a/patching/README.md
+++ b/patching/README.md
@@ -1,0 +1,97 @@
+# Patching Sample
+
+This sample shows how to safely alter a workflow using `patched` and `deprecate_patch` in stages.
+
+To run, first see [README.md](../README.md) for prerequisites. Then follow the patching stages below.
+
+### Stage 1 - Initial code
+
+This stage is for existing running workflows. To simulate our initial workflow, run the worker in a separate terminal:
+
+    poetry run python worker.py --workflow initial
+
+Now we can start this workflow:
+
+    poetry run python starter.py --start-workflow initial-workflow-id
+
+This will output "Started workflow with ID initial-workflow-id and ...". Now query this workflow:
+
+    poetry run python starter.py --query-workflow initial-workflow-id
+
+This will output "Query result for ID initial-workflow-id: pre-patch".
+
+### Stage 2 - Patch the workflow
+
+This stage is for needing to run old and new workflows at the same time. To simulate our patched workflow, stop the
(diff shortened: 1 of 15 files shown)


patching/starter.py
```diff
@@ -0,0 +1,34 @@
...
+
+# Since it's just used for typing purposes, it doesn't matter which one we
+# import
+from workflow_1_initial import MyWorkflow
+
+
+async def main():
+    parser = argparse.ArgumentParser(description="Run worker")
+    parser.add_argument("--start-workflow", help="Start workflow with this ID")
+    parser.add_argument("--query-workflow", help="Query workflow with this ID")
```
bergundy, line 14: nit: I'd make these subcommands:
https://docs.python.org/3/library/argparse.html#sub-commands

I'd also encourage using a nicer argument parser like [click](https://click.palletsprojects.com/en/8.1.x/) for samples where we don't care too much about the additional dependency.

patching/README.md
```diff
@@ -0,0 +1,97 @@
+# Patching Sample
+
+This sample shows how to safely alter a workflow using `patched` and `deprecate_patch` in stages.
```
bergundy, line 3: Why did you choose this workflow as the sample for patches?
It's not especially practical and doesn't show how patches help making workflow changes that would otherwise break determinism, changing a query result can be done without patching.

cretz: I didn't really choose one, I just wanted to focus on the patch part without creating activities or timers.

cretz: I have updated the sample to use activities instead. I was gonna demonstrate the non-determinism in README, but I can't until I upgrade Python to a core version with https://github.com/temporalio/sdk-core/pull/475.

bergundy (approved):

PR #69, opened 2023-05-09
Remove TODOs for issues 200/201

r00ta:
## What was changed
The issues 200 and 201 are fixed so this PR aims to remove the TODOs

## Why?
Ready to process TODOs

## Checklist

1. Closes 

2. How was this tested:
manually

3. Any docs updates needed?


cretz (approved):
Thanks! Will merge if/when CI passes

PR #71, opened 2023-05-25
fix: update queue generation to be one sticky queue per worker host

### LINKED ISSUE #70: [Bug] Activity sticky queue example written incorrectly
### Describe the bug

The sample has a select-task-queue activity that uses a random set of task queues. But it's supposed to return a fixed task queue for the host its on. See the TypeScript sample it references.

Oracen:
## What was changed
Sticky task queues changed from having multiple sticky queues per worker to single sticky queue per worker. 

## Why?
Issue opened, need to standardise with the Typescript sticky queues example.

## Checklist
1. Closes #70

2. How was this tested:
Commands in `README.md` run, tests run via `poe test`.

3. Any docs updates needed?
Minor changes to descriptions.


cretz:
Bypassing CLA, was signed in #45

Issue #72, opened 2023-06-09
[Feature Request] Replace thrift jaeger exporter with otlp


romank0:
### Is your feature request related to a problem? Please describe.

The opentelementy example uses thrift jaeger exporter instead of more generic `opentelemetry-exporter-otlp-proto-grpc`. This was done due to [this issue](https://github.com/open-telemetry/opentelemetry-python/issues/2880) in `opentelemetry-exporter-otlp-proto-grpc` that has beed fixed already.

### Describe the solution you'd like

Make use of the `opentelemetry-exporter-otlp-proto-grpc`

### Additional context

N/A

PR #73, opened 2023-06-09
replaces thrift collector with otlp collector

romank0:
## What was changed
thrift jaeger spans exporter was replaced by OTLP exporter in opentelemetry example.

## Why?
OTLP exporter is more generic as it uses open protocol supported by mutlipe providers.

## Checklist

1. Closes https://github.com/temporalio/samples-python/issues/72

2. How was this tested:
I've run manually the example of how to use opentelemetry from README.

3. Any docs updates needed?
Yes. README for opentelemetry example is updated.


cretz (approved):
Great! Glad to see this works with OTel now. Will approve for CI. Click CLA link and agree to that thing when you can and I'll merge.

PR #78, opened 2023-07-18
Worker Versioning sample

Sushisource:
Draft until python sdk is released with this feature


cretz (commented):
LGTM, but will also refrain from approving until SDK updated

cretz (approved):

PR #79, opened 2023-07-24
restruct sentry proj

bkodes:
The code in /sentry currently does not run.
```
Cannot access http.client.IncompleteRead.__mro_entries__ from inside a workflow. If this is code from a module not used in a workflow or known to only be us
ed deterministically from a workflow, mark the import as pass through.
```

Separating the workflow code into its own file resolves the issue

## What was changed

Moved greetings workflow to workflows.py and import workflow within worker

## Why?

So others can reference this sample and have no issue running it without alterations being made

## Checklist

1. Closes 

2. How was this tested:

Followed run instructions in README.md. Outputs:
```
cmd: poetry run python starter.py
output: Workflow result: Hello, World!
```

3. Any docs updates needed?

No

### DIFF
--- a/sentry/worker.py
+++ b/sentry/worker.py
@@ -5,35 +5,11 @@ from dataclasses import dataclass
 from datetime import timedelta
 
 import sentry_sdk
-from temporalio import activity, workflow
+from sentry.workflows import compose_greeting, GreetingWorkflow
 from temporalio.client import Client
 from temporalio.worker import Worker
 
 from sentry.interceptor import SentryInterceptor
-
-
-@dataclass
-class ComposeGreetingInput:
-    greeting: str
-    name: str
-
-
-@activity.defn
-async def compose_greeting(input: ComposeGreetingInput) -> str:
-    activity.logger.info("Running activity with parameter %s" % input)
-    return f"{input.greeting}, {input.name}!"
-
-
-@workflow.defn
-class GreetingWorkflow:
-    @workflow.run
-    async def run(self, name: str) -> str:
-        workflow.logger.info("Running workflow with parameter %s" % name)
-        return await workflow.execute_activity(
-            compose_greeting,
-            ComposeGreetingInput("Hello", name),
-            start_to_close_timeout=timedelta(seconds=10),
-        )
 
 
 async def main():
--- a/sentry/worker.py
+++ b/sentry/worker.py
@@ -5,7 +5,8 @@ from dataclasses import dataclass
 from datetime import timedelta
 
 import sentry_sdk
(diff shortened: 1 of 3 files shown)


sentry/workflows.py
```diff
@@ -0,0 +1,26 @@
...
+    return f"{input.greeting}, {input.name}!"
+
+
+@workflow.defn
+class GreetingWorkflow:
+    @workflow.run
+    async def run(self, name: str) -> str:
+        workflow.logger.info("Running workflow with parameter %s" % name)
+        return await workflow.execute_activity(
+            compose_greeting,
```
cretz, line 23: This is referencing `compose_greeting` which doesn't exist

cretz (commented):
Thanks! Moving workflows to a separate file is definitely a good thing these days. Just need to get that activity back in there and I think we'll be good.

sentry/workflows.py
```diff
@@ -0,0 +1,16 @@
+from temporalio import activity, workflow
+from dataclasses import dataclass
+from datetime import timedelta
+from sentry.activities import ComposeGreetingInput, compose_greeting
```
cretz, line 4:
```suggestion
# Import our activity and param class, passing it through the sandbox
with workflow.unsafe.imports_passed_through():
    from sentry.activities import ComposeGreetingInput, compose_greeting
```

You want to pass this through the sandbox, you don't want to reload it everytime

sentry/workflows.py
cretz, line 1: I think you should update the `starter.py` to reference the workflow from this file and not through the `worker.py` file

sentry/worker.py
```diff
@@ -5,37 +5,13 @@
 from datetime import timedelta
 
 import sentry_sdk
-from temporalio import activity, workflow
+from sentry.workflows import compose_greeting, GreetingWorkflow
```
cretz, line 8:
```suggestion
from sentry.activities import compose_greeting
from sentry.workflows import GreetingWorkflow
```

Should import from the canonical place the thing is defined IMO

cretz (commented):
Please make sure you run `poe lint` before pushing (`poe format` will fix the formatting issues)

brianstrauch:
The restructure you proposed here has since landed in `main` — the sentry sample now has `sentry/workflow.py` and `sentry/activity.py` as separate modules imported by `sentry/worker.py`, and the sandbox import error is handled explicitly with `workflow.unsafe.imports_passed_through()`. Most of that came in via the Sentry V2 SDK rework (#140) and follow-up refactors.

Since the change is no longer applicable to the current sample, I'm closing this out.

PR #84, opened 2023-10-11
Gevent sample

cretz:
## What was changed

Create a gevent sample showing how to use Temporal client, worker, workflows, and activities in a gevent-patched environment

### DIFF
--- a/gevent_async/executor.py
+++ b/gevent_async/executor.py
@@ -1,0 +1,41 @@
+import functools
+from concurrent.futures import Future
+from typing import Callable, TypeVar
+
+from gevent import threadpool
+from typing_extensions import ParamSpec
+
+T = TypeVar("T")
+P = ParamSpec("P")
+
+
+class GeventExecutor(threadpool.ThreadPoolExecutor):
+    def submit(
+        self, fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs
+    ) -> Future[T]:
+        # Gevent's returned futures do not map well to Python futures, so we
+        # must translate. We can't just use set_result/set_exception because
+        # done callbacks are not always called in gevent's case and it doesn't
+        # seem to support cancel, so we instead wrap the caller function.
+        python_fut: Future[T] = Future()
+
+        @functools.wraps(fn)
+        def wrapper(*w_args: P.args, **w_kwargs: P.kwargs) -> None:
+            try:
+                result = fn(*w_args, **w_kwargs)
+                # Swallow InvalidStateError in case Python future was cancelled
+                try:
+                    python_fut.set_result(result)
+                except:
+                    pass
+            except Exception as exc:
+                # Swallow InvalidStateError in case Python future was cancelled
+                try:
+                    python_fut.set_exception(exc)
+                except:
+                    pass
+
+        # Submit our wrapper to gevent
+        super().submit(wrapper, *args, **kwargs)
+        # Return Python future to user
+        return python_fut
--- a/gevent_async/starter.py
+++ b/gevent_async/starter.py
@@ -1,0 +1,40 @@
+# Init gevent
+from gevent import monkey
+
+monkey.patch_all()
+
+import asyncio
+import logging
+
+from temporalio.client import Client
+
+from gevent_async import workflow
+from gevent_async.executor import GeventExecutor
+
+
+def main():
+    logging.basicConfig(level=logging.INFO)
+
+    # Create single-worker gevent executor and run asyncio.run(async_main()) in
+    # it, waiting for result. This executor cannot be used for anything else in
+    # Temporal, it is just a single thread for running asyncio.
+    with GeventExecutor(max_workers=1) as executor:
+        executor.submit(asyncio.run, async_main()).result()
+
+
+async def async_main():
+    # Connect client
+    client = await Client.connect("localhost:7233")
+
+    # Run workflow
+    result = await client.execute_workflow(
+        workflow.GreetingWorkflow.run,
+        "Temporal",
+        id=f"gevent_async-workflow-id",
+        task_queue="gevent_async-task-queue",
+    )
+    logging.info(f"Workflow result: {result}")
+
+
+if __name__ == "__main__":
+    main()
--- a/gevent_async/starter.py
+++ b/gevent_async/starter.py
@@ -30,7 +30,7 @@ async def async_main():
     result = await client.execute_workflow(
         workflow.GreetingWorkflow.run,
         "Temporal",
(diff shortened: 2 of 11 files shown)


gevent_async/executor.py
```diff
@@ -0,0 +1,41 @@
...
+                # Swallow InvalidStateError in case Python future was cancelled
+                try:
+                    python_fut.set_result(result)
+                except:
+                    pass
+            except Exception as exc:
+                # Swallow InvalidStateError in case Python future was cancelled
+                try:
+                    python_fut.set_exception(exc)
+                except:
```
dandavison, line 35: Since this might end up in production code, I suggest catching the specific exception that you indicate in the comment.

gevent_async/starter.py
```diff
@@ -0,0 +1,40 @@
...
+
+async def async_main():
+    # Connect client
+    client = await Client.connect("localhost:7233")
+
+    # Run workflow
+    result = await client.execute_workflow(
+        workflow.GreetingWorkflow.run,
+        "Temporal",
+        id=f"gevent_async-workflow-id",
```
dandavison, line 33: nit: pointless `f`

gevent_async/executor.py
cretz: Technically we support 3.7+ and this class didn't get added until 3.8 and we try to write our code for all versions. Having said that, gevent only supports 3.8+ too so it may be worth changing, we'll see.

gevent_async/starter.py
cretz: Thanks!

josh-berry:
Tried running this on Mac and the worker hangs on `Client.connect("localhost:7233")` (with the latest released CLI). Looking into why.

cretz:
This does not work on Linux, still investigating...

gevent_async/starter.py
cretz: Fixed.

cretz:
Fixed, please review

gevent_async/executor.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class GeventExecutor(threadpool.ThreadPoolExecutor):
+    def submit(
+        self, fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs
+    ) -> Future[T]:
+        # Gevent's returned futures do not map well to Python futures, so we
+        # must translate. We can't just use set_result/set_exception because
+        # done callbacks are not always called in gevent's case and it doesn't
+        # seem to support cancel, so we instead wrap the caller function.
```
josh-berry, line 19: [non-blocking/optional] I think it might be clearer to explicitly state "Gevent's returned futures do not implement all of the things expected by `concurrent.futures.Future`" rather than saying they "do not map well". I wasn't sure what the latter meant, so I wound up reading gevent code to figure out what this meant.

josh-berry (approved):
Just one suggestion on the documentation. Otherwise LGTM.

gevent_async/executor.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import Callable, TypeVar
+
+from gevent import threadpool
+from typing_extensions import ParamSpec
+
+T = TypeVar("T")
+P = ParamSpec("P")
+
+
+class GeventExecutor(threadpool.ThreadPoolExecutor):
```
abeland, line 12: Hi @cretz, I believe gevent.threadpoo.ThreadPoolExecutor is the *unpatched* thread pool, so really this PR is just using native threads.

cf. https://www.gevent.org/api/gevent.threadpool.html#gevent.threadpool.ThreadPoolExecutor :

> A version of [concurrent.futures.ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) that always uses native threads, even when threading is monkey-patched.

I verified by running this code locally with the patched `concurrent.futures.ThreadPoolExecutor` and it throws this error, as seen before in our explorations:

```
[~/temporalio_gevent/src] ∴ ~/.local/bin/poetry run python -m gevent_async.worker
INFO:root:Worker started, ctrl+c to exit
INFO:temporalio.workflow:Running workflow with parameter Temporal ({'attempt': 1, 'namespace': 'default', 'run_id': '543fef67-e25d-4bd8-926a-cca7bc78fd91', 'task_queue': 'gevent_async-task-queue', 'workflow_id': 'gevent_async-workflow-id', 'workflow_type': 'GreetingWorkflow'})
ERROR:temporalio.worker._workflow:Failed handling activation on workflow with run ID 543fef67-e25d-4bd8-926a-cca7bc78fd91
Traceback (most recent call last):
  File "/home/discord/.cache/pypoetry/virtualenvs/temporalio-gevent-EEP-UPYz-py3.10/lib/python3.10/site-packages/temporalio/worker/_workflow.py", line 208, in _handle_activation
    completion = await asyncio.wait_for(
  File "/usr/lib/python3.10/asyncio/tasks.py", line 405, in wait_for
    loop = events.get_running_loop()
RuntimeError: no running event loop
```

cretz: Exactly. We are _intentionally_ using native threads so that gevent doesn't try to mess with them so you can have asyncio running in a native thread and gevent doing other things elsewhere. The whole crux of the original issue is that you couldn't even use Temporal and gevent in the same process because gevent tries to take over, but this allows them to share the process space.

If we need further investigation on using greenlets maybe from activities or something, we can look into that separately.

alecgorge: @cretz please correct me if my understanding is wrong: if this is intentionally using native threads, it is creating a new thread for each activity it runs [on line 39](https://github.com/temporalio/samples-python/pull/84/files#diff-e15a336a556fdd3034e734ee18a80f0d12180da39ac407b655174ff085eef8e9R39) despite the comment on line 38 saying that it will "[s]ubmit our wrapper to gevent". this means that all the code executed in activities needs to be threadsafe.

our codebase isn't threadsafe (pretty common for gevent codebases i think) but we are still giving this a go to see if we can get basic execution working.

alecgorge:
The examples shown here never worked for us because (a) our app wasn't threadsafe and (b) we didn't want to be spawning N separate native threads.

Here is a solution I put together that should only have 2 threads: one aux thread to handle asyncio things and one gevent main thread that processes activities: https://gist.github.com/alecgorge/63e52b764b10adc46d61b1773f9a2085

It doesn't handle a mixture of async and sync (gevent) activities, but that is fine for our purposes.

cretz:
Great! Yeah, that's the crux of it, is that asyncio and gevent threads don't mix. The _entire_ asyncio event loop running in a single separate thread from gevent is the best way however you can accomplish that.

PR #91, opened 2023-11-03
Adds ruff as a drop in replacement for black

rachfop:
## What was changed
Replaces Black with Ruff formatter.

Recently, ruff added the ability to format:
https://docs.astral.sh/ruff/formatter/

## Known deviations: 

https://docs.astral.sh/ruff/formatter/black/#newlines-are-inserted-after-all-class-docstrings

## Removed isort:
Added a skip isort in the .giitignore file:
https://pycqa.github.io/isort/docs/configuration/action_comments.html
> Ruff respects isort's [action comments](https://pycqa.github.io/isort/docs/configuration/action_comments.html) (# isort: skip_file, # isort: on, # isort: off, # isort: skip, and # isort: split), which enable selectively enabling and disabling import sorting for blocks of code and other inline configuration.

Ruff uses isort's `profile = black` by default:
https://docs.astral.sh/ruff/faq/#how-does-ruffs-import-sorting-compare-to-isort>
> Ruff's import sorting is intended to be near-equivalent to isort's when using isort's profile = "black".

## Why?

Faster 🐎 
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz:
Nice! I'd like to see about doing this to our SDK repo too. Can it replace `isort` too?

rachfop:
Good call. Removed `isort` and updated the description of this pr with how it works.

cretz:
@rachfop - Not understanding exactly. So does this provide all the import sorting benefit of isort? If we're removing isort, why have a comment in the gitignore relating to the tool we don't use/support anymore?

rachfop:
Ruff supports isort's action comments, so we can remove isort from the pyproject.tom, but use isort's `skip_file` and ruff wil respect that.
I only included that because it was a part of the original pyproject.toml sorting logic.

cretz (commented):
LGTM, thanks!

rachfop:
Resolved merge!

brianstrauch:
Closing as superseded — the repo has since moved to Ruff for both formatting and import sorting (see the `format`/`lint` tasks in `pyproject.toml`), and `black`/`isort` are gone. The lockfile portion of this diff also no longer applies since the migration to uv (#170).

PR #96, opened 2023-11-22
Use temporalio.testing.ActivityEnvironment

rachfop:
## What was changed
Some sample showing activity testing without the workflow side.

## Why?

For users who might want to test Activities.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):
Thanks!

PR #99, opened 2023-12-12
Adds Workflow Update to Samples

rachfop:
## What was changed

Adds the Workflow Update to the sample's repo.

## Why?

Because users will want to see how to use the new Update feature with Python.
Gracefully shut down the worker too.

## Checklist

1. Closes 

2. How was this tested:

You need the Temporal Python 1.4 or above.
Only works on the local dev server with the config setup.

3. Any docs updates needed?

Once this gets approved - I'll add something similar to the docs (and incorporate any additional feedback).


README.md
```diff
@@ -66,6 +66,7 @@ Some examples require extra dependencies. See each sample's directory for specif
 * [sentry](sentry) - Report errors to Sentry.
 * [worker_specific_task_queues](worker_specific_task_queues) - Use unique task queues to ensure activities run on specific workers.
 * [worker_versioning](worker_versioning) - Use the Worker Versioning feature to more easily version your workflows & other code.
+* [workflow_update](workflow_update) - Send a request to and a response from a client to a workflow execution.
```
cretz, line 69: I think such a simple feature sample may deserve to be with the others in `hello`. I understand there's temporary extra README info while it is experimental (we can add to bottom of hello README), but otherwise I think of this as a basic demonstration of a basic feature like signals, queries, etc.

(granted would love to do #67 and not have those be single files, but that can come separate/later for all hello samples)

workflow_update/starter.py
```diff
@@ -0,0 +1,39 @@
+import asyncio
+
+from temporalio.client import Client, WorkflowHandle
+
+from workflow_update import HelloWorldWorkflow
+
+
+class WorkflowRunner:
```
cretz, line 8: I don't think there's much value in this separate class, just put the `start_workflow` in the `run_workflow` method

cretz (approved):
This is great, thanks!

cretz:
Looks like you may need to do a `poe format`. Also, `hello/hello_update.py` may be a better name than `hello/hello_workflow_update.py` to match signals, but this is fine too.

PR #100, opened 2023-12-15
Adds LangChain translation example

rachfop:
## What was changed

Add an example of LangChain.

## Checklist

1. Closes 

I need help with Poetry and a Pydantic Converter.

I tried setting the dev dependencies like this, so I wouldn't have to change the top-level version - but that didn't work.
So I just overwrote the Python dependcy.

```
dependencies = { langchain = { version = "^0.0.350", python = ">=3.8.1,<4.0" }, openai = { version = "^1.4.0", python = ">=3.8,<4.0" }, fastapi = { version="^0.105.0", python = ">=3.8,<4.0"}, uvicorn = { version = "^0.24.0.post1", extras = ["standard"], python = ">=3.8,<4.0" } }
```

While this code works, it probably needs a Pydantic Convert for V2 to properly serialize. 

3. How was this tested:

4. Any docs updates needed?


pyproject.toml
```diff
@@ -16,7 +16,7 @@ packages = [
 "Bug Tracker" = "https://github.com/temporalio/samples-python/issues"
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = ">=3.8.1,<4.0"
```
cretz, line 19: I think we should have a PR first that just drops 3.7 support and moves CI from 3.7 + 3.11 to 3.8 + 3.12. We did this for the primary SDK at https://github.com/temporalio/sdk-python/pull/422. If you want to make that part of this issue, I'd be ok with that too (but change the title too).

pyproject.toml
```diff
@@ -41,6 +41,10 @@ dependencies = { cryptography = "^38.0.1", aiohttp = "^3.8.1" }
 optional = true
 dependencies = { gevent = { version = "^23.9.1", python = ">=3.8" } }
 
+[tool.poetry.group.langchain]
+optional = true
+dependencies = { langchain =  "^0.0.350", openai = "^1.4.0", fastapi = "^0.105.0", uvicorn = { version = "^0.24.0.post1", extras = ["standard"]} }
```
cretz, line 46: If it's easier/clearer, you can change this to:

```suggestion
[tool.poetry.group.langchain.dependencies]
langchain =  "^0.0.350"
openai = "^1.4.0"
fastapi = "^0.105.0"
uvicorn = { version = "^0.24.0.post1", extras = ["standard"]}
```

langchain/starter.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@app.post("/translate")
+async def translate(phrase: str, language: str):
+    client = app.state.temporal_client
+    try:
+        result = await client.execute_workflow(
+            LangChainWorkflow.run,
+            TranslateParams(phrase, language),
+            id=f"langchain-translation-{language.lower()}-{phrase.replace(' ', '-')}",
```
cretz, line 26: I think you should just make a unique ID each time in this case

pyproject.toml
cretz: Already done as part of #101. This PR needs to be updated.

PR #105, opened 2024-02-19
Modernize payload converter sample

drewhoskins:
## What was changed

In the custom converter sample, use the more modern, non-initialized style of fetching the default payload converter.

## Why?

Acting on @cretz's TODO.

## Checklist

1. Closes 

2. How was this tested:

`poetry run pytest ./tests/custom_converter/workflow_test.py`

3. Any docs updates needed?


cretz (approved):
Thanks! Will merge if/when CI completes (we may have a hold-up with a MyPy check we need to fix elsewhere)

PR #107, opened 2024-02-27
require frozenlist to update version

axfelix:
add `frozenlist = "^1.4.0"` to dev-dependencies since this was trying to install an older version that wasn't building right under Python 3.12 / MSVC 2022


cretz:
Can you give more details? Our CI passes on Windows 3.12, here's the latest: https://github.com/temporalio/samples-python/actions/runs/7977279807/job/21779816371. Is there some test we could write that would make CI fail if this fix wasn't in place?

axfelix:
Yeah, it was triggering this PEP 517 build error in a relatively stock environment:

https://stackoverflow.com/questions/64038673/could-not-build-wheels-for-which-use-pep-517-and-cannot-be-installed-directly

My knowledge of Python bootstrapping is somewhat out of date -- it's possible our build system is doing something different than my environment to make the build pass -- but updating the dependency to a supported version that has wheels available seemed like a decent idea regardless.

cretz:
> it's possible our build system is doing something different than my environment to make the build pass -- but updating the dependency to a supported version that has wheels available seemed like a decent idea regardless.

Sure, but I want to make sure we catch future issues like this too, but if our build environment is not representative in some way I think we should fix.

> Also -- this was only occurring when doing poetry install --with encryption as in https://github.com/temporalio/samples-python/tree/main/encryption, which we aren't testing currently.

:+1: So I wonder if we should add `--with encryption` to our `poetry install` set in `.github/workflows/ci.yml` that would help catch this

axfelix:
Done! I strongly suspect that the `ci.yml` change would not work without the `pyproject.toml` change -- let me know if you'd like me to verify this further before merging.

cretz (approved):
> I strongly suspect that the ci.yml change would not work without the pyproject.toml change -- let me know if you'd like me to verify this further before merging.

Good enough for me! Merge at will.

PR #110, opened 2024-04-18
add export proto to parquet example

alice-yin:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


export_proto_to_parquet/LICENSE
cretz, line 1: Can you remove this file

export_proto_to_parquet/README.md
cretz, line 1: Can we change this sample to `cloud_export_to_parquet` so it's clear, unlike every other sample we right, that it's cloud specific. The `proto` part isn't important I don't think.

export_proto_to_parquet/README.md
```diff
@@ -0,0 +1,59 @@
...
+python -m pip install pandas
+python -m pip install pyarrow
+python -m pip install boto3
+```
+
+
+Run the workflow:
+
+```bash
+python run_workflow.py
```
cretz, line 34:
```suggestion
python create_schedule.py
```

Can we change the filename to be a bit clearer on what's happening in the script?

export_proto_to_parquet/run_workflow.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+    try:
+        await client.create_schedule(
+            "hourly-proto-to-parquet-wf-schedule",
+            Schedule(
+                action=ScheduleActionStartWorkflow(
+                    ProtoToParquet.run,
+                    wf_input,
+                    id=f"{WORKFLOW_ID_PREFIX}-{datetime.now()}",
```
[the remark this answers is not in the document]
alice-yin: So I have given some screenshot using this type of workflow id. So would like to keep this format

export_proto_to_parquet/workflows.py
```diff
@@ -0,0 +1,67 @@
...
+        # Read from export S3 bucket and given at least 2 hour delay to ensure the file has been uploaded
+        read_time = workflow.now() - timedelta(hours=workflow_input.num_delay_hour)
+        common_path = f"{workflow_input.namespace}/{read_time.year}/{read_time.month:02}/{read_time.day:02}/{read_time.hour:02}/00"
+        path = f"temporal-workflow-history/export/{common_path}"
+        get_object_keys_input = GetObjectKeysActivityInput(
+            workflow_input.export_s3_bucket, path
+        )
+
+        # Read Input File
+        object_keys_output = await workflow.execute_activity_method(
```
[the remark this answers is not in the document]
alice-yin: Should not too big. These are file names in S3.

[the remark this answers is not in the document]
alice-yin: So the benefit I think we could do multi thread for data_trans_and_land activity. That's taking the majority of time right now.

export_proto_to_parquet/run_workflow.py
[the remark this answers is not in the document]
cretz: I think we may need to be able to improve the sample in the future regardless of how it's referenced in a post. I think the screenshot can differ, but up to you, no strong need to change this now.

export_proto_to_parquet/workflows.py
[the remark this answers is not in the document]
cretz: But this isn't doing anything concurrently today, it's one at a time. But I agree in general.

alice-yin:
Did quite a bit refactoring on this.

cloud_export_to_parquet/README.md
```diff
@@ -0,0 +1,23 @@
+# Cloud Export to parquet sample
+
+This is an example workflow to convert exported file from proto to parquet file. The workflow is an hourly schedule. 
+
+Please make sure your python is 3.9 above. For this sample, run:
+
+    poetry install
```
cretz, line 7: This should be something like `poetry install --with cloud_export_to_parquet`

cloud_export_to_parquet/data_trans_activities.py
```diff
@@ -0,0 +1,129 @@
...
+    return object_keys
+
+
+@activity.defn
+async def data_trans_and_land(activity_input: DataTransAndLandActivityInput) -> str:
+    """Function that convert proto to parquet and save to S3."""
+    key = activity_input.object_key
+    data = await get_data_from_object_key(activity_input.export_s3_bucket, key)
+    activity.logger.info("Convert proto to parquet for file: %s", key)
+    parquet_data = convert_proto_to_parquet_flatten(data)
```
cretz, line 53: I am concerned this is CPU bound work, is it expensive (i.e. can it take a lot of time blocking the entire asyncio event loop)?

I think `def` activities would be better than `async def` activities for code that does data translation and such.

cloud_export_to_parquet/data_trans_activities.py
```diff
@@ -0,0 +1,129 @@
...
+            row_flatten.insert(0, "WorkflowId", worfkow_id)
+            row_flatten.insert(1, "RunId", run_id)
+            rows_flatten.append(row_flatten)
+    df_flatten = pd.concat(rows_flatten, ignore_index=True)
+    return df_flatten
+
+
+async def save_to_sink(data: pd.DataFrame, s3_bucket: str, write_path: str) -> str:
+    """Function that save object to s3 bucket."""
+    write_bytes = data.to_parquet(None, compression="snappy", index=False)
```
cretz, line 116: Is this an expensive operation that blocks the asyncio event loop?

cloud_export_to_parquet/pyproject.toml
cretz, line 1: Samples do not need their own pyproject.toml. You can just add a group to the top-level one

cloud_export_to_parquet/README.md
alice-yin: So, as the project depends on different python. Thus, I can't merge them in the same pyproject.toml. The current pyproject.toml needs to use python <= 3.8 but pandas need to be use python >= 3.9. I tried use pandas that could use python 3.8, however there are other dependency issue as well.

cloud_export_to_parquet/pyproject.toml
alice-yin: same comment as above. Not be able to use the same pyproject.toml.

cloud_export_to_parquet/data_trans_activities.py
alice-yin: So while we do convert_proto_to_parquet_flatten, reading from s3 and writing to s3 are also I/O  happy operations, it would be benefit if we still keep it as async. I could improve the code a bit. 

Also I thought it's always recommended to use async def for activities? Am I wrong on this?

cloud_export_to_parquet/README.md
cretz: I think it should be ok to say in this sample's README that this sample requires >= 3.9. Is that possible or are you hitting other issues?

cloud_export_to_parquet/data_trans_activities.py
cretz: >  it would be benefit if we still keep it as async

I don't think so in this case, I think normal `def` would be best

> Also I thought it's always recommended to use async def for activities? Am I wrong on this?

No, actually the opposite, you only use `async def` if you're sure you aren't going to block the thread. We had that async-default stance originally but we've changed documentation and courses and such to encourage just `def` instead.

> I was planning to add the following logic in workflows.py

No need, workflow is good IMO

cloud_export_to_parquet/README.md
alice-yin: So that's why I keep a separate poetry file inside the folder. Thus, it could build independently with the root folder.

cloud_export_to_parquet/data_trans_activities.py
alice-yin: Changed all functions to sync.

cloud_export_to_parquet/README.md
cretz: I think you can make the Python version higher on just your dependencies in your group, see https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies (if our poetry version too old we can update). So you can say "You need Python 3.9+ and then run `poetry install --with cloud_export_to_parquet`".

cloud_export_to_parquet/run_worker.py
```diff
@@ -0,0 +1,34 @@
...
+    # Run the worker
+    worker: Worker = Worker(
+        client,
+        task_queue="DATA_TRANSFORMATION_TASK_QUEUE",
+        workflows=[ProtoToParquet],
+        activities=[get_object_keys, data_trans_and_land],
+        workflow_runner=SandboxedWorkflowRunner(
+            restrictions=SandboxRestrictions.default.with_passthrough_modules("boto3")
+        ),
+        activity_executor=ThreadPoolExecutor(5),
```
cretz, line 28:
```suggestion
        activity_executor=ThreadPoolExecutor(100),
```

Or should make the `max_concurrent_activities` smaller, your choice, but thread pool in theory should not be less than `max_concurrent_activities`.

cloud_export_to_parquet/README.md
alice-yin: Updated it.

cretz (approved):
Great! This is a great looking sample that is nice and simple to read. Merge whenever you'd like.

poetry.lock
cretz, line 1: It looks like you inadvertently updated a lot of dependencies instead of just the ones for your project (this is probably causing whatever MyPy to break). Usually we're ok with updating dependencies, but we may want it in a separate PR.

PR #113, opened 2024-04-25
batch_daily example

hellais:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/batch_daily/workflows.py
+++ b/batch_daily/workflows.py
@@ -1,0 +1,84 @@
+from datetime import datetime, timedelta
+
+from dataclasses import dataclass
+
+from temporalio import workflow
+from temporalio.common import RetryPolicy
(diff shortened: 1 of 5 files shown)


batch_daily/workflows.py
```diff
@@ -0,0 +1,106 @@
...
+
+        record_uri_list = await workflow.execute_activity(
+            list_records,
+            list_records_input,
+            start_to_close_timeout=timedelta(minutes=5),
+        )
+
+        task_list = []
+        async with asyncio.TaskGroup() as tg:
+            for key in record_uri_list:
```
hellais, line 57: This is no good as it runs the risk of spamming the event history log. It should probably work similarly to the java example where it uses continue_as_new to reset the event history.

PR #114, opened 2024-05-22
Message processing design patterns

dandavison:


signals_and_updates/serialized_handling_of_n_messages.py
cretz, line 1: Here are three alternatives that are untested, I just typed real quick, but the ideas are there

## Same as original design (CAN w/ update failures)

```python
from collections import deque
from dataclasses import dataclass, field
from typing import Optional

from temporalio import workflow

# !!!
# This is a copy of the original design, but it is flawed because it fails
# updates to the caller on CAN
# !!!


@dataclass
class MessageProcessorInput:
    pending_tasks: list[str] = field(default_factory=list)


class UpdateTask:
    def __init__(self, arg: str) -> None:
        self.arg = arg
        self.result: Optional[str] = None
        self.returned = False


@workflow.defn
class MessageProcessor:
    def __init__(self) -> None:
        self.inited = False

    @workflow.run
    async def run(self, input: MessageProcessorInput) -> None:
        # Startup
        self.queue = deque([UpdateTask(arg) for arg in input.pending_tasks])
        self.inited = True

        # Process until CAN is needed
        while not workflow.info().is_continue_as_new_suggested():
            await workflow.wait_condition(lambda: len(self.queue) > 0)
            await self.process_task(self.queue.popleft())

        # CAN knowing that pending updates will fail
        workflow.continue_as_new(
            MessageProcessorInput(pending_tasks=[task.arg for task in self.queue])
        )

    @workflow.update
    async def do_task(self, arg: str) -> str:
        # Wait for primary workflow to be inited
        await workflow.wait_condition(lambda: self.inited)

        # Add task and wait on result
        task = UpdateTask(arg)
        try:
            self.queue.append(task)
            await workflow.wait_condition(lambda: task.result is not None)
            assert task.result
            return task.result
        finally:
            task.returned = True

    async def process_task(self, task: UpdateTask) -> None:
        task.result = await workflow.execute_activity(
            "some_activity", task.arg, start_to_close_timeout=timedelta(seconds=10)
        )
        await workflow.wait_condition(lambda: task.returned)
```

## Update doesn't wait

```python
from collections import deque

from temporalio import workflow

# !!!
# This version does not make update wait on completion
# !!!


@workflow.defn
class MessageProcessor:
    def __init__(self) -> None:
        self.inited = False

    @workflow.run
    async def run(self, queue: deque[str]) -> None:
        self.queue = queue
        self.inited = True
        # Process until CAN is needed
        while not workflow.info().is_continue_as_new_suggested():
            await workflow.wait_condition(lambda: len(self.queue) > 0)
            await self.process_task(self.queue.popleft())

        workflow.continue_as_new(self.queue)

    @workflow.update
    async def do_task(self, arg: str) -> None:
        # Wait for primary workflow to be inited
        await workflow.wait_condition(lambda: self.inited)
        # Put on queue and complete update
        self.queue.append(arg)

    async def process_task(self, arg: str) -> None:
        await workflow.execute_activity(
            "some_activity", arg, start_to_close_timeout=timedelta(seconds=10)
        )
```

## Update must complete with result

```python
from collections import deque
from typing import Optional

from temporalio import workflow

# !!!
# This version requires update to complete with result and won't CAN until after
# everything is done
# !!!


class UpdateTask:
    def __init__(self, arg: str) -> None:
        self.arg = arg
        self.result: Optional[str] = None
        self.returned = False


@workflow.defn
class MessageProcessor:
    def __init__(self) -> None:
        self.queue: deque[UpdateTask] = deque()

    @workflow.run
    async def run(self) -> None:
        # Process until CAN is needed and the queue is empty
        while not workflow.info().is_continue_as_new_suggested() or len(self.queue) > 0:
            await workflow.wait_condition(lambda: len(self.queue) > 0)
            await self.process_task(self.queue.popleft())

        # CAN knowing the queue is empty
        workflow.continue_as_new()

    @workflow.update
    async def do_task(self, arg: str) -> str:
        # Add task and wait on result
        task = UpdateTask(arg)
        try:
            self.queue.append(task)
            await workflow.wait_condition(lambda: task.result is not None)
            assert task.result
            return task.result
        finally:
            task.returned = True

    async def process_task(self, task: UpdateTask) -> None:
        task.result = await workflow.execute_activity(
            "some_activity", task.arg, start_to_close_timeout=timedelta(seconds=10)
        )
        await workflow.wait_condition(lambda: task.returned)
```

antlai-temporal: In the third case, "Update must complete with result", how to guarantee that CAN eventually happens if updates keep on coming...?

cretz: > In the third case, "Update must complete with result", how to guarantee that CAN eventually happens if updates keep on coming...? 

You can't. You'd only choose this use case if you were ensured a period of idleness between tasks. You can definitely have a version that rejects updates when `is_continue_as_new_suggested()` is present if you wanted.

PR #115, opened 2024-05-22
Don't mutate class attribute

dandavison:
- Don't mutate class attribute
- Comment out unused imports

### DIFF
--- a/encryption/codec.py
+++ b/encryption/codec.py
@@ -1,4 +1,3 @@
-import base64
 import os
 from typing import Iterable, List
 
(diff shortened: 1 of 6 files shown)


cretz (approved):

cretz:
@dandavison - the CI issue is caused by `macos-latest` now pointing to `macos-14` which is arm instead of intel. The fix is probably to use `actions/setup-python@v4` instead of `actions/setup-python@v1`, set `macos-intel` and `macos-arm` as `macos-12` and `macos-14` respectively, and disable 3.8 for `macos-arm`. See https://github.com/temporalio/sdk-python/pull/506. Can be done in this PR or in a separate one your choice (or work can just be put on the backlog if time is limited).

dandavison:
> the CI issue is caused...

Thanks! Opened https://github.com/temporalio/samples-python/issues/117 for that.

PR #116, opened 2024-05-23
Bedrock sample

brendan-myers:
## What was changed

Add examples for Amazon Bedrock.

## Why?

These are/will be referenced in an upcoming Temporal blog.

## Checklist

1. Closes 

n/a

3. How was this tested:

Tested as per README in the individual Bedrock sample folders. To test you will need a set of AWS credentials, and permissions to access Bedrock.

4. Any docs updates needed?

Yes. Description and link to samples added to README.


bedrock/3_EntityWorkflow/end_chat.py
```diff
@@ -0,0 +1,20 @@
+import asyncio
+import sys
+from temporalio.client import Client
+from workflows import EntityBedrockWorkflow
+
+
+async def main(prompt):
```
cretz, line 7: Unnecessary prompt parameter

bedrock/2_SignalsAndQueries/get_history.py
```diff
@@ -0,0 +1,25 @@
+import asyncio
+from temporalio.client import Client
+from workflows import SignalQueryBedrockWorkflow
+
+
+async def main():
+    client = await Client.connect("localhost:7233")
+    workflow_id = "simple-bedrock-workflow-1"
```
cretz, line 8: Consider using different workflow identifiers for different samples

bedrock/3_EntityWorkflow/end_chat.py
```diff
@@ -0,0 +1,20 @@
...
+    client = await Client.connect("localhost:7233")
+
+    workflow_id = "simple-bedrock-workflow-1"
+
+    handle = client.get_workflow_handle(workflow_id=workflow_id)
+
+    # sends a signal to the workflow (and starts it if needed)
```
cretz, line 14:
```suggestion
    # sends a signal to the workflow
```

bedrock/3_EntityWorkflow/run_worker.py
```diff
@@ -0,0 +1,33 @@
...
+import logging
+from temporalio.client import Client
+from temporalio.worker import Worker
+
+from activities import prompt_bedrock
+from workflows import EntityBedrockWorkflow
+
+
+async def main():
+    # Create client connected to server at the given address
```
cretz, line 12: Inconsistent with whether you capitalize your comments and inconsistent with whether you add comments above client connect

bedrock/3_EntityWorkflow/workflows.py
```diff
@@ -0,0 +1,174 @@
...
+        )
+        self.prompt_queue: Deque[str] = deque()
+        self.conversation_summary: Optional[str] = None
+        self.continue_as_new_per_turns: int = 6
+        self.chat_ended: bool = False
+
+    @workflow.run
+    async def run(
+        self,
+        params: BedrockParams = None,
```
cretz, line 33: This will not pass type checking. You can't assign `None` to a non-optional parameter.

bedrock/3_EntityWorkflow/workflows.py
```diff
@@ -0,0 +1,174 @@
...
+    ) -> str:
+
+        if params and params.conversation_summary:
+            self.conversation_history.append(
+                ("conversation_summary", params.conversation_summary)
+            )
+
+        if params and params.prompt_queue:
+            for prompt in params.prompt_queue:
+                self.prompt_queue.append(prompt)
```
cretz, line 43:
```suggestion
            self.prompt_queue.extend(params.prompt_queue)
```

bedrock/3_EntityWorkflow/workflows.py
```diff
@@ -0,0 +1,174 @@
...
+@workflow.defn
+class EntityBedrockWorkflow:
+    def __init__(self) -> None:
+        # List to store prompt history
+        self.conversation_history: List[Tuple[str, str]] = (
+            []
+        )
+        self.prompt_queue: Deque[str] = deque()
+        self.conversation_summary: Optional[str] = None
+        self.continue_as_new_per_turns: int = 6
```
cretz, line 27: We usually encourage using `workflow.info().is_continue_as_new_suggested()`, but don't have to

bedrock/3_EntityWorkflow/workflows.py
```diff
@@ -0,0 +1,174 @@
...
+
+            workflow.logger.info(f"{response}")
+
+            # Append the response to the conversation history
+            self.conversation_history.append(("response", response))
+
+            # summarize the conversation to date using Amazon Bedrock
+            # uses start_activity with a callback
+            # so it doesn't block new messages being sent to Amazon Bedrock
+            summary_activity_task = workflow.start_activity(
```
cretz, line 97: So you are summarizing after every user prompt? If I send 20 user prompts, that's 20 summary calls? If this is a requirement, arguably no value in having it be separate activities then (the same activity can issue single prompt and request summary w/ full history).

bedrock/3_EntityWorkflow/workflows.py
```diff
@@ -0,0 +1,174 @@
...
+                prompt_bedrock,
+                self.prompt_summary_from_history(),
+                schedule_to_close_timeout=timedelta(seconds=20),
+            )
+            summary_activity_task.add_done_callback(self.summary_complete)
```
cretz, line 102: This is racy. What if a summary task from an early iteration of the loop finishes after a summary task from a later iteration? You should make summary occur at the end IMO or on user request.

But if you do want to leave it like this, you should make sure only one runs at a time and it'd probably be clearer to use an `async def` to do it and store the result instead of a done callback (and use `asyncio.create_task` to run it in the background if needed). This approach does not properly account for failures of the activity (will throw in the callback which goes nowhere).

bedrock/3_EntityWorkflow/workflows.py
```diff
@@ -0,0 +1,174 @@
...
+        while True:
+            workflow.logger.info("Waiting for prompts...")
+
+            # Wait for a chat message (signal) or timeout
+            await workflow.wait_condition(
+                lambda: bool(self.prompt_queue) or self.chat_ended
+            )
+
+            # if end chat signal was sent
+            if self.chat_ended:
```
cretz, line 56: This will lose prompts if they sent a prompt signal _and_ a chat end signal

bedrock/3_EntityWorkflow/workflows.py
```diff
@@ -0,0 +1,174 @@
...
+                # chat to summarize
+                if summary_activity_task is not None:
+                    # ensure conversation summary task has finished
+                    # before closing the workflow (avoid race)
+                    await workflow.wait_condition(
+                        lambda: summary_activity_task.done()
+                    )
+                else:
+                    # conversation history from previous workflow
+                    self.conversation_summary = params.conversation_summary
```
cretz, line 67: Can put this in the initialization code at the top of this method

bedrock/3_EntityWorkflow/workflows.py
```diff
@@ -0,0 +1,174 @@
...
+            # if end chat signal was sent
+            if self.chat_ended:
+                # the workflow might be continued as new without any
+                # chat to summarize
+                if summary_activity_task is not None:
+                    # ensure conversation summary task has finished
+                    # before closing the workflow (avoid race)
+                    await workflow.wait_condition(
+                        lambda: summary_activity_task.done()
+                    )
```
cretz, line 64: Can also just `await summary_activity_task` (but recognize exceptions can occur)

cretz, line 64: This can cause signal loss if a signal comes in while waiting on this summary

bedrock/3_EntityWorkflow/workflows.py
```diff
@@ -0,0 +1,174 @@
...
+
+            # We summarize the chat to date and use that as input to the
+            # new workflow
+            if len(self.conversation_history) >= self.continue_as_new_per_turns:
+                # ensure conversation summary task has finished
+                # before continuing as new
+                if summary_activity_task is not None:
+                    await workflow.wait_condition(
+                        lambda: summary_activity_task.done()
+                    )
```
cretz, line 117: Causes signal loss as mentioned in other spots

bedrock/2_SignalsAndQueries/workflows.py
```diff
@@ -0,0 +1,110 @@
...
+                await workflow.wait_condition(
+                    lambda:
+                        bool(self.prompt_queue),
+                        timeout=timedelta(minutes=inactivity_timeout_minutes),
+                )
+            # if timeout was reached
+            except asyncio.TimeoutError:
+                workflow.logger.info("Chat closed due to inactivity")
+                # End the workflow
+                break
```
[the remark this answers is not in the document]
brendan-myers: If we've reached timeout, and are running final activities to close out the workflow, are signals that are received but not processed considered lost? What should be the expected behaviour?

bedrock/3_EntityWorkflow/workflows.py
brendan-myers: Good catch. Have flipped this so if both are received then all prompts are processed first.

bedrock/3_EntityWorkflow/workflows.py
brendan-myers: Have fixed this to be consistent with the signals and queries sample - where summarize is only called at conversation end. Although in this case it's also summarized on CAN.

bedrock/3_EntityWorkflow/workflows.py
brendan-myers: Where did we land on this @steveandroulakis? For context, this is currently done this way so a/ developers can easily trigger CAN, b/ the token input limit to the Bedrock model may be lower than the Temporal history limit.

steveandroulakis: I think manually setting the limit is preferred for those reasons. You might want to have a couple of # comments saying "set this lower to trigger CAN (for learning)" or something. And also even note the suggest..() function exists?

bedrock/3_EntityWorkflow/workflows.py
steveandroulakis: This is a good point, thanks Chad. I might refactor this so the summary is triggered only when needed (before CAN / and at the end).

bedrock/3_EntityWorkflow/workflows.py
steveandroulakis: @cretz when you say signal loss, do you mean this code would behave like so:
* workflow is waiting on summary before continuing as new
* a signalABC comes in while waiting on summary
* summary finishes and workflow does CAN without handling signalABC (signal is effectively dropped)

How does one manage this in Python? I know in Go that we can explicitly drain signals but in the Python docs I don't see anything about draining/signal loss (and thought that Rust Core SDKs ones don't need this).

bedrock/2_SignalsAndQueries/workflows.py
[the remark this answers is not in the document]
cretz: > are signals that are received but not processed considered lost?

Usually, since they might as well not be received if you never do anything with them. Maybe that is the behavior you want, but it's not the common behavior. Usually people want to process every signal they get. Closing something due to inactivity when there is activity right at the end is not usually good behavior, you should consider signals coming in anytime before actual close as the workflow no longer being "inactive".

bedrock/3_EntityWorkflow/workflows.py
cretz: I admit not digging into the sample, but token limit of some system should not affect CAN. But yes, if you are not using `is_continue_as_new_suggested()` because you want to CAN earlier just for the sample, ok, but yes it's worth mentioning. For most users it is not worth CAN'ing so frequently.

bedrock/3_EntityWorkflow/workflows.py
cretz: > How does one manage this in Python? I know in Go that we can explicitly drain signals but in the Python docs I don't see anything about draining/signal loss

You do similar to Go. You don't need explicit "draining" in Python because we guarantee your handler code is _always_ called which we don't in Go. But from a business-logic POV, don't CAN while there are signals to be processed.

bedrock/1_BasicBedrockWorkflow/workflows.py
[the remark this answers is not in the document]
brendan-myers: Discussed with @steveandroulakis - we'll leave out tests for this example. There are examples ([1](https://github.com/temporalio/samples-python/blob/main/tests/hello/hello_query_test.py), [2](https://github.com/temporalio/samples-python/blob/main/tests/hello/hello_signal_test.py), [3](https://github.com/temporalio/samples-python/blob/main/tests/hello/hello_activity_test.py)) that cover the basics of how one would write tests for this.

bedrock/2_SignalsAndQueries/workflows.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(self) -> None:
+        # List to store prompt history
+        self.conversation_history: List[Tuple[str, str]] = (
+            []
+        )
+        self.prompt_queue: Deque[str] = deque()
+        self.conversation_summary: str = ""
+
+    @workflow.run
+    async def run(self, inactivity_timeout_minutes: int) -> str:
```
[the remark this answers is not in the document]
brendan-myers: The entity example now uses data classes. Will leave the signals and queries example unchanged, as it's only a single input, so between the two examples it nicely demonstrates the different ways to pass inputs to a workflow.

bedrock/basic/README.md
```diff
@@ -0,0 +1,10 @@
+# Basic Amazon Bedrock workflow
+
+A basic Bedrock workflow. Starts a workflow with a prompt, generates a response and ends the workflow.
+
+To run, first see [README.md](../README.md) for prerequisites. Then, run the following from this directory:
```
cretz, line 5:
```suggestion
To run, first see [README.md](../../README.md) for prerequisites. Then, run the following from this directory:
```

Do you want to point them to the base README for prereqs?

bedrock/basic/workflows.py
```diff
@@ -0,0 +1,24 @@
...
+
+        workflow.logger.info("Prompt: %s" % prompt)
+
+        response = await workflow.execute_activity_method(
+            BedrockActivities.prompt_bedrock,
+            prompt,
+            schedule_to_close_timeout=timedelta(seconds=20),
+        )
+
+        workflow.logger.info("Response: %s" % response)
```
cretz, line 22: It can be a bit confusing to have some output be logs and some be `print` (but is probably ok)

bedrock/basic/send_message.py
```diff
@@ -0,0 +1,29 @@
+import asyncio
+import sys
+
+from temporalio.client import Client
+from workflows import BasicBedrockWorkflow
+
+
+async def main(prompt):
```
cretz, line 8:
```suggestion
async def main(prompt: str) -> str:
```

Usually best to type-hint things to/from workflow

bedrock/entity/end_chat.py
```diff
@@ -0,0 +1,22 @@
...
+from workflows import EntityBedrockWorkflow
+
+
+async def main():
+    # Create client connected to server at the given address
+    client = await Client.connect("localhost:7233")
+
+    workflow_id = "entity-bedrock-workflow"
+
+    handle = client.get_workflow_handle(workflow_id)
```
cretz, line 14:
```suggestion
    handle = client.get_workflow_handle_for(EntityBedrockWorkflow.run, workflow_id)
```

Not required, but this is a bit more type-safe

bedrock/entity/send_message.py
```diff
@@ -0,0 +1,30 @@
...
+async def main(prompt):
+    # Create client connected to server at the given address
+    client = await Client.connect("localhost:7233")
+
+    workflow_id = "entity-bedrock-workflow"
+
+    # Sends a signal to the workflow (and starts it if needed)
+    await client.start_workflow(
+        EntityBedrockWorkflow.run,
+        arg=BedrockParams(None, None),
```
cretz, line 17:
```suggestion
        BedrockParams(None, None),
```

Shouldn't need a kwarg for this

bedrock/entity/workflows.py
```diff
@@ -0,0 +1,169 @@
...
+
+        while True:
+            workflow.logger.info("Waiting for prompts...")
+
+            # Wait for a chat message (signal) or timeout
+            await workflow.wait_condition(
+                lambda: bool(self.prompt_queue) or self.chat_ended
+            )
+
+            if bool(self.prompt_queue):
```
cretz, line 53:
```suggestion
            if self.prompt_queue:
```

Shouldn't need to convert to `bool` here

bedrock/entity/workflows.py
```diff
@@ -0,0 +1,169 @@
...
+                )
+
+                return f"{self.conversation_history}"
+
+    @workflow.signal
+    async def user_prompt(self, prompt: str) -> None:
+        # Chat ended but the workflow is waiting for a chat summary to be generated
+        if self.chat_ended:
+            workflow.logger.warn(f"Message dropped due to chat closed: {prompt}")
+            return
```
cretz, line 128: This is an ok approach, though often something like prompt + end will come in at what appears to be "same time" from separate clients, but server happens to get end chat first which means data loss. But that's probably ok for this sample. If workflow update were stable we could return back an error instead of warn.

bedrock/entity/workflows.py
```diff
@@ -0,0 +1,169 @@
...
+    def prompt_summary_from_history(self) -> str:
+        history_string = self.format_history()
+        return (
+            "Here is the conversation history between a user and a chatbot: "
+            + f"{history_string}  -- Please produce a two sentence summary of "
+            + "this conversation."
+        )
+
+    # Callback -- save the latest conversation history once generated
+    def summary_complete(self, task: asyncio.Task) -> None:
```
cretz, line 168: Is this method needed?

bedrock/signals_and_queries/workflows.py
```diff
@@ -0,0 +1,109 @@
...
+                )
+            # If timeout was reached
+            except asyncio.TimeoutError:
+                self.chat_timeout = True
+                workflow.logger.info("Chat closed due to inactivity")
+                # End the workflow
+                break
+
+            # Fetch next user prompt and add to conversation history
+            prompt = self.prompt_queue.popleft()
```
cretz, line 43: You should process the whole queue here instead of just re-running the loop for each item. Usually re-running a loop would be no big deal, but this starts a timer each time for the timeout. So if you have 100 signals that come in at the same time, this code will start 100 timers as opposed to probably just 1. Arguably we should not even start timers if the lambda evaluates to true immediately, but we do.

cretz (commented):
Looks good, only minor things (and many (all?) can be ignored)

bedrock/entity/workflows.py
steveandroulakis: Great catch, thank you

bedrock/basic/README.md
cretz: Note @steveandroulakis this applies to all three sub-READMEs and not just this one

steveandroulakis:
Ah, of course! Will resolve

Steve Androulakis (andrew-lah-kis)
Solutions Architect
***@***.***

steveandroulakis:
Updated all 3 sub-READMEs to be specific about 2 sets of pre-requisites (samples-python and bedrock)

bedrock/entity/workflows.py
brendan-myers: Agree, but I also think that for this example it's okay. For production we'd want to be more careful about how we handle this case, but the tradeoff is additional code/complexity which is probably too much for this sample.

brendan-myers:
@cretz all changes made

cretz (approved):
Looks good, thanks! Would have liked some tests, but not all of our samples have tests. Still, if y'all have the time would strongly encourage it so this doesn't just break one day without us knowing.

cretz:
GH actions broken right now, but once CI passes feel free to merge

PR #122, opened 2024-06-07
fix missing sleep to properly poll inside activity

mnichols:
Current `frequent` polling implementation doesnt ever sleep after the mock service replies with a "service is down" Exception. This is supposed to be demonstrating polling _inside_ the activity so let's add back the sleep and handle heartbeat error upon cancellation correctly.

### DIFF
--- a/polling/frequent/activities.py
+++ b/polling/frequent/activities.py
@@ -24,7 +24,7 @@ async def compose_greeting(input: ComposeGreetingInput) -> str:
                 return result
             except Exception as e:
                 # swallow exception since service is down
-                activity.logger.error(e)
+                activity.logger.debug("Failed, trying again shortly", exc_info=True)
 
             activity.heartbeat("Invoking activity")
             await asyncio.sleep(1)
--- a/polling/frequent/activities.py
+++ b/polling/frequent/activities.py
@@ -33,4 +33,3 @@ async def compose_greeting(input: ComposeGreetingInput) -> str:
             # if you need to clean up you can catch this.
             # Here we are just reraising the exception
             raise
-
--- a/polling/frequent/activities.py
+++ b/polling/frequent/activities.py
@@ -19,7 +19,7 @@ async def compose_greeting(input: ComposeGreetingInput) -> str:
     while True:
         try:
             try:
-                result = test_service.get_service_result(input)
+                result = await test_service.get_service_result(input)
                 activity.logger.info(f"Exiting activity ${result}")
                 return result
             except Exception as e:
--- a/polling/frequent/activities.py
+++ b/polling/frequent/activities.py
@@ -18,18 +18,19 @@ async def compose_greeting(input: ComposeGreetingInput) -> str:
     test_service = TestService()
     while True:
         try:
-            result = test_service.get_service_result(input)
-            activity.logger.info(f"Exiting activity ${result}")
-            return result
-        except Exception as e:
-            # swallow exception since service is down
-            activity.logger.error(e)
+            try:
+                result = test_service.get_service_result(input)
+                activity.logger.info(f"Exiting activity ${result}")
+                return result
+            except Exception as e:
+                # swallow exception since service is down
+                activity.logger.error(e)
 
-        try:
             activity.heartbeat("Invoking activity")
-        except asyncio.CancelledError as exception:
+            await asyncio.sleep(1)
+        except asyncio.CancelledError:
             # activity was either cancelled or workflow was completed or worker shut down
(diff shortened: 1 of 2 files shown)


polling/frequent/activities.py
```diff
@@ -17,6 +20,16 @@ async def compose_greeting(input: ComposeGreetingInput) -> str:
     while True:
         try:
             result = test_service.get_service_result(input)
+            print(f"exiting activity ${result}")
```
cretz, line 23: I wonder if we should either remove the print statements or use a logger

```suggestion
            activity.logger.info(f"Exiting activity ${result}")
```

polling/frequent/activities.py
```diff
@@ -17,6 +20,16 @@ async def compose_greeting(input: ComposeGreetingInput) -> str:
...
-        except Exception:
+        except Exception as e:
+            # swallow exception since service is down
+            print(e)
+
+        try:
             activity.heartbeat("Invoking activity")
+        except CancelledError as exception:
+            # activity was either cancelled or workflow was completed or worker shut down
+            raise exception
```
cretz, line 33: Heartbeat does not throw. Cancellation in Python occurs at the asyncio level so it will throw whatever awaits (so probably on sleep or RPC call). And it uses asyncio cancellation, so it'd be `asyncio.CancelledError` not the Temporal one. Also, there is no value in catching an exception just to re-raise (unless it's just for the doc), and if you do, just call `raise` no need to pass anything.

polling/frequent/activities.py
mnichols: fixed!

polling/frequent/activities.py
mnichols: I think the doc nudge to handle cancellation is merited here as it is in the samples-java repo as well. I know it has helped me to remember to deal with cancellation correctly.

polling/frequent/activities.py
```diff
@@ -17,6 +19,17 @@ async def compose_greeting(input: ComposeGreetingInput) -> str:
...
+        except Exception as e:
+            # swallow exception since service is down
+            activity.logger.error(e)
+
+        try:
             activity.heartbeat("Invoking activity")
+        except asyncio.CancelledError as exception:
+            # activity was either cancelled or workflow was completed or worker shut down
+            # if you need to clean up you can catch this. Here we are just reraising exception
+            raise
```
cretz, line 33: Heartbeat does not raise an exception. This exception would occur from the get service result or sleep call

polling/frequent/activities.py
```diff
@@ -17,6 +19,17 @@ async def compose_greeting(input: ComposeGreetingInput) -> str:
     while True:
         try:
             result = test_service.get_service_result(input)
```
cretz, line 21: I think we need to change `get_service_result` to `async def` and change everywhere that calls it to use `await` if they don't already

polling/frequent/activities.py
```diff
@@ -17,6 +19,17 @@ async def compose_greeting(input: ComposeGreetingInput) -> str:
...
+            activity.logger.info(f"Exiting activity ${result}")
             return result
-        except Exception:
+        except Exception as e:
+            # swallow exception since service is down
+            activity.logger.error(e)
```
cretz, line 26: I don't think you should log the exception at the error level if it's expected to keep happening, consider:

```suggestion
            activity.logger.debug("Failed, trying again shortly", exc_info=True)
```

polling/frequent/activities.py
mnichols: OK that confuses me. I reimplemented with whole thing in CancelledError because I still think it matters

polling/frequent/activities.py
mnichols: the service today is just a synchronous mock of a thing. maybe do this improvement as a follow up enhancement since this is just fixing sample as stated?

cretz: The problem is that the infrequent equivalent of this calling the same mock does to `await` (I admittedly didn't review this code well enough apparently when it was originally added). I think it's important for _this_ example to `await` the call too because it's important to show that this is an await point from which a cancellation can throw. If you don't want to change the signature on the other side to `async def`, sure we can open a GH issue to do that.

mnichols: understood...fixed!

mnichols: I also made the signature of test_service explicitly async

cretz (approved):
Looks great, thanks! May need to run a `poe format` on the repo to get it to pass CI.

cretz:
Looks great, merge away

Issue #49, opened 2023-01-25
[Feature Request] Guide against single-file structure


lorensr:
### Is your feature request related to a problem? Please describe.

Given this single-file pattern of running a Workflow inside a Worker:

https://github.com/temporalio/samples-python/blob/7b3944926c3743bc0dcb3b781d8cc64e0330bac4/hello/hello_activity.py#L47-L64

there is potential (and one past instance I know of) for people learning based off of samples to think that in order to run a Workflow, you need to run it inside an `async with Worker`. When they try to develop an application based on that paradigm, they run into a number of issues.

### Describe the solution you'd like

Make it clear that when developing, we recommend:

- running a worker separately from workflows
- defaulting to a single worker for all workflows, not one worker per workflow

We could make it clear by structuring all samples that way. OTOH I like the brevity of the single file samples. Another possibility is changing the comment. The drawbacks to that are:

- some people don't read comments
- the comment might not be clear, unless you point to a multi-file sample, what exactly we recommend doing

Comment is currently:

```
# While the worker is running, use the client to run the workflow and
# print out its result. Note, in many production setups, the client
# would be in a completely separate process from the worker.
```

Perhaps could be:

```
# Use the client to run the workflow and print out its result. 
# NOTE: when developing, we recommend starting out running a 
# single worker that has all your workflow and activities and
# running client code in a separate process. See, for example,
# how the encryption sample has a separate files to run:
# worker.py runs the worker and starter.py uses the client.
```

cretz:
I support any changes you suggest to this effect. We initially wanted single file examples for clarity. I support breaking them into multiple files, changing the comment, or maybe even the best option - same file but different paths based on arg.

mbernier:
I was thinking about suggesting a change where the first example is a single file and has a comment that says something like the above to say "Yo, this is a single file for a simple hello world. All other examples break this into multiple files, you should not build your app in the single file way". 

Something extremely clear and to the point. 

The other nice thing about multi-file examples, is that a README could be included to explain what is going on with that specific example and why it is included.

cretz:
> you should not build your app in the single file way

FWIW this is subjective based on use case. Many people may have the same app starting the workflow as running it. It's just that most people choose to start somewhere separate than they run.

> The other nice thing about multi-file examples, is that a README could be included to explain what is going on with that specific example and why it is included.

[...]

PR #123, opened 2024-06-19
Workflow Update and Signal handlers concurrency sample

drewhoskins-temporal:
## What was changed

Added a ClusterManager sample that shows off `workflow.wait_condition` in handlers as well as the use of a mutex to guarantee atomicity.

## Why?

As part of our effort to teach users about interleaving of blocking signal and update handlers, as well as about a workflow's reentrancy model in general, we are producing samples.

## Checklist

1. Closes 

2. How was this tested:

`poetry run pytest tests/updates_and_signals/atomic_message_handlers_test.py`

3. Any docs updates needed?


update_and_signal_handlers/atomic_message_handlers.py
cretz, line 1: README should be updated referencing this sample

cretz, line 1: We consider it bad practice to put non-workflow code with workflow code and so we don't do it in samples except for the `hello` ones (which we may change see https://github.com/temporalio/samples-python/issues/49 and https://github.com/temporalio/samples-python/issues/67). Users have done bad things combining code since entire workflow files run in a sandbox including all non-workflow code/imports. Can we break this out to separate files like the other non-hello samples?

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+from datetime import timedelta
+import logging
+from typing import Dict, List, Optional
+import uuid
+
+from temporalio import activity, common, workflow
+from temporalio.client import Client, WorkflowHandle
+from temporalio.worker import Worker
+
+# This samples shows off the key concurrent programming primitives for Workflows, especially
```
cretz, line 11:
```suggestion
# This sample shows off the key concurrent programming primitives for Workflows, especially 
```

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+# useful for workflows that handle signals and updates.
+
+#   - Makes signal and update handlers only operate when the workflow is within a certain state 
+#     (here between cluster_started and cluster_shutdown) using workflow.wait_condition.
+#   - Signal and update handlers can block and their actions can be interleaved with one another and with the main workflow.
+#     Here, we use a lock to protect shared state from interleaved access.
+#   - Running start_workflow with an initializer signal that you want to run before anything else.
+#
+@activity.defn
+async def allocate_nodes_to_job(nodes: List[int], task_name: str) -> List[int]:
```
cretz, line 21: Return doesn't match type hint (here and elsewhere)

cretz, line 21: We usually discourage multi-param activities/workflows in favor of single dataclass instances with multiple fields

update_and_signal_handlers/atomic_message_handlers.py
cretz, line 1: Please add a README for this sample similar to other samples

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+        self.cluster_started = False
+        self.cluster_shutdown = False
+        # Protects workflow state from interleaved access
+        self.nodes_lock = asyncio.Lock()
+        self.max_assigned_nodes = 0
+
+    @workflow.signal
+    async def start_cluster(self):
+        self.cluster_started = True
+        self.nodes : Dict[Optional[str]] = dict([(k, None) for k in range(25)])
```
cretz, line 54: Dicts require types for keys and values and usually you'd use a dictionary comprehension (e.g. `{k: None for k in range(25)}`)

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+        self.cluster_shutdown = True
+        workflow.logger.info("Cluster shut down")
+
+    @workflow.update
+    async def allocate_n_nodes_to_job(self, task_name: str, num_nodes: int, ) -> List[int]:
+        await workflow.wait_condition(lambda: self.cluster_started)
+        assert not self.cluster_shutdown
+
+        await self.nodes_lock.acquire()
+        try:
```
cretz, line 69: Consider using `async with self.nodes_lock:`, I think it's a bit clearer

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+        workflow.logger.info("Cluster started")
+
+    @workflow.signal
+    async def shutdown_cluster(self):
+        await workflow.wait_condition(lambda: self.cluster_started)
+        self.cluster_shutdown = True
+        workflow.logger.info("Cluster shut down")
+
+    @workflow.update
+    async def allocate_n_nodes_to_job(self, task_name: str, num_nodes: int, ) -> List[int]:
```
cretz, line 64: Unnecessarily trailing param comma (can use `poe format` to format code)

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+        await workflow.wait_condition(lambda: self.cluster_started)
+        assert not self.cluster_shutdown
```
cretz, line 66: This would fail the workflow task causing the workflow to hang. Is this intentional? If not, consider using an exception that would fail the update.

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+            self.nodes_lock.release()
+
+    @workflow.run
+    async def run(self):
+        await workflow.wait_condition(lambda: self.cluster_started)
+
+        # Perform health checks at intervals
+        while True:
+            try:
+                await workflow.wait_condition(lambda: self.cluster_shutdown, timeout=timedelta(seconds=1))
```
cretz, line 130: We try to avoid such frequent timers in loops in workflows usually (this can blow up histories), so even in a sample it may be encouraging bad behavior

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+
+    async def perform_health_checks(self):
+        await self.nodes_lock.acquire()
+        try:
+            assigned_nodes = [k for k, v in self.nodes.items() if v is not None]
+            # This await would be dangerous without nodes_lock because it yields control and allows interleaving.
+            bad_nodes = await workflow.execute_activity(find_bad_nodes, assigned_nodes, start_to_close_timeout=timedelta(seconds=10))
+            for node in bad_nodes:
+                self.nodes[node] = "BAD!"
+            self.num_assigned_nodes = len(assigned_nodes)
```
cretz, line 119: What is this instance variable used for?

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+        for node in nodes_to_free:
+            self.nodes[node] = None
+
+    async def perform_health_checks(self):
+        await self.nodes_lock.acquire()
+        try:
+            assigned_nodes = [k for k, v in self.nodes.items() if v is not None]
+            # This await would be dangerous without nodes_lock because it yields control and allows interleaving.
+            bad_nodes = await workflow.execute_activity(find_bad_nodes, assigned_nodes, start_to_close_timeout=timedelta(seconds=10))
```
cretz, line 116: May want to clarify in a comment here and on `_allocate_nodes_to_job` what will happen if these activities fail and retry forever (or never start for whatever reason) to confirm it is intentional (lock will never get unlocked)

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+            for node in bad_nodes:
+                self.nodes[node] = "BAD!"
+            self.num_assigned_nodes = len(assigned_nodes)
+        finally:
+            self.nodes_lock.release()
+
+    @workflow.run
+    async def run(self):
```
cretz, line 124: Would encourage explicit return type hints on workflow functions

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+        await workflow.wait_condition(lambda: self.cluster_started)
+
+        # Perform health checks at intervals
+        while True:
+            try:
+                await workflow.wait_condition(lambda: self.cluster_shutdown, timeout=timedelta(seconds=1))
+            except asyncio.TimeoutError:
+                pass
+            if self.cluster_shutdown:
+                break
```
cretz, line 134: Could just put `break` after the `wait_condition` line inside the `try`

update_and_signal_handlers/atomic_message_handlers.py
```diff
@@ -0,0 +1,179 @@
...
+    await asyncio.gather(*deletion_updates)
+        
+    await wf.signal(ClusterManager.shutdown_cluster)
+
+async def main():
+    client = await Client.connect("localhost:7233")
+
+    async with Worker(
+        client,
+        task_queue="tq",
```
cretz, line 159: To prevent clashing, in samples we try to name the task queue after the sample

tests/update_and_signal_handlers/atomic_message_handlers_test.py
```diff
@@ -0,0 +1,26 @@
...
+from temporalio import workflow, common
+from temporalio.client import Client
+from temporalio.worker import Worker
+from update_and_signal_handlers.atomic_message_handlers import ClusterManager, allocate_nodes_to_job, deallocate_nodes_for_job, do_cluster_lifecycle, find_bad_nodes
+
+
+async def test_atomic_message_handlers(client: Client):
+    async with Worker(
+        client,
+        task_queue="tq",
```
cretz, line 12: Would suggest unique task queues in tests

tests/update_and_signal_handlers/atomic_message_handlers_test.py
```diff
@@ -0,0 +1,26 @@
...
+        workflows=[ClusterManager],
+        activities=[allocate_nodes_to_job, deallocate_nodes_for_job, find_bad_nodes],
+    ):
+        cluster_manager_handle = await client.start_workflow(
+            ClusterManager.run,
+            id=f"ClusterManager-{uuid.uuid4()}",
+            task_queue="tq",
+            id_reuse_policy=common.WorkflowIDReusePolicy.TERMINATE_IF_RUNNING,
```
cretz, line 20: Should not be necessary in tests (tests should be isolated where they shouldn't have to worry about other things that could be running)

update_and_signal_handlers/atomic_message_handlers.py
drewhoskins-temporal: Yes changing the hello samples would be great as that's what I pattern-matched off of.

cretz: :+1: I think there is some resistance to doing so, but yeah in the meantime I think matching the other whole-directory samples will work best here.

cretz:
@drewhoskins-temporal - let me know when ready for re-review

updates_and_signals/atomic_message_handlers/README.md
cretz, line 1: I think this can just be at a top-level directory of `atomic_message_handlers`, no need to nest an extra directory deep

cretz, line 1: The primary README at the root of this repo should be updated to reference this sample

updates_and_signals/atomic_message_handlers/activities.py
```diff
@@ -0,0 +1,45 @@
+import asyncio
+from dataclasses import dataclass
+from typing import List
+
+from temporalio import activity
+
+
+@dataclass(kw_only=True)
```
cretz, line 8:
```suggestion
@dataclass
```

Probably not needed, but no big deal

updates_and_signals/atomic_message_handlers/activities.py
```diff
@@ -0,0 +1,45 @@
...
+class AllocateNodesToJobInput:
+    nodes: List[str]
+    task_name: str
+
+
+@activity.defn
+async def allocate_nodes_to_job(input: AllocateNodesToJobInput) -> List[str]:
```
cretz, line 15: This return type hint seems invalid (same with some other functions)

updates_and_signals/atomic_message_handlers/starter.py
```diff
@@ -0,0 +1,70 @@
...
+    # Connect to Temporal
+    client = await Client.connect("localhost:7233")
+
+    cluster_manager_handle = await client.start_workflow(
+        ClusterManagerWorkflow.run,
+        ClusterManagerInput(test_continue_as_new=should_test_continue_as_new),
+        id=f"ClusterManagerWorkflow-{uuid.uuid4()}",
+        task_queue="atomic-message-handlers-task-queue",
+        id_reuse_policy=common.WorkflowIDReusePolicy.TERMINATE_IF_RUNNING,
+        start_signal="start_cluster",
```
cretz, line 49: While I understand this is demonstrating handlers, arguably for users there is not much value of combining these two options together. If you know you always want to do something at the start of the workflow you could call it at the start of the workflow (e.g. when there is no state). No problem with it being here though, may just be a bit confusing.

cretz (commented):
Mostly LGTM, only minor things

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -0,0 +1,198 @@
...
+    async def _deallocate_nodes_for_job(self, nodes_to_free: List[str], task_name: str):
+        await workflow.execute_activity(
+            deallocate_nodes_for_job,
+            DeallocateNodesForJobInput(nodes=nodes_to_free, task_name=task_name),
+            start_to_close_timeout=timedelta(seconds=10),
+        )
+        for node in nodes_to_free:
+            self.state.nodes[node] = None
+
+    def get_assigned_nodes(self) -> List[str]:
```
cretz, line 124: Any concerns that this list may be mutated during iteration by a running update since it is not called under lock from `run`?

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -0,0 +1,198 @@
...
+@dataclass(kw_only=True)
+class ClusterManagerInput:
+    state: Optional[ClusterManagerState] = None
+    test_continue_as_new: bool = False
+
+
+@dataclass
+class ClusterManagerResult:
+    max_assigned_nodes: int
+    num_assigned_nodes: int
```
cretz, line 43: Porting this to .NET and not sure there is value storing this field on "state" (and it's built out of lock, so it's a bit confusing)

updates_and_signals/atomic_message_handlers/starter.py
```diff
@@ -0,0 +1,70 @@
...
+    ClusterManagerWorkflow,
+)
+
+
+async def do_cluster_lifecycle(wf: WorkflowHandle, delay_seconds: Optional[int] = None):
+    allocation_updates = []
+    for i in range(6):
+        allocation_updates.append(
+            wf.execute_update(
+                ClusterManagerWorkflow.allocate_n_nodes_to_job, args=[f"task-{i}", 2]
```
cretz, line 21: I think we want to discourage multiple arguments to things (workflows, activities, signals, queries, updates, etc)

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -0,0 +1,198 @@
...
+
+    @workflow.update
+    async def allocate_n_nodes_to_job(
+        self,
+        task_name: str,
+        num_nodes: int,
+    ) -> List[str]:
+        await workflow.wait_condition(lambda: self.state.cluster_started)
+        if self.state.cluster_shutdown:
+            raise RuntimeError(
```
cretz, line 77: This (and the `ValueError` below) are task failures. You may want to use `ApplicationError`.

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -0,0 +1,198 @@
...
+
+# In workflows that continue-as-new, it's convenient to store all your state in one serializable structure
+# to make it easier to pass between runs
+@dataclass(kw_only=True)
+class ClusterManagerState:
+    cluster_started: bool = False
+    cluster_shutdown: bool = False
+    nodes: Optional[Dict[str, Optional[str]]] = None
+    max_assigned_nodes: int = 0
+    num_assigned_nodes: int = 0
```
cretz, line 31: Porting this to .NET and not sure there is value storing this num-node field on "state" (and it's built out of lock, so it's a bit confusing)

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -0,0 +1,198 @@
...
+
+    @workflow.update
+    async def delete_job(self, task_name: str) -> str:
+        await workflow.wait_condition(lambda: self.state.cluster_started)
+        assert not self.state.cluster_shutdown
+        async with self.nodes_lock:
+            nodes_to_free = [k for k, v in self.state.nodes.items() if v == task_name]
+            # This await would be dangerous without nodes_lock because it yields control and allows interleaving.
+            await self._deallocate_nodes_for_job(nodes_to_free, task_name)
+        return "Done"
```
cretz, line 113: Probably don't need to return a value from this update

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -0,0 +1,198 @@
...
+            AllocateNodesToJobInput(nodes=assigned_nodes, task_name=task_name),
+            start_to_close_timeout=timedelta(seconds=10),
+        )
+        for node in assigned_nodes:
+            self.state.nodes[node] = task_name
+
+    @workflow.update
+    async def delete_job(self, task_name: str) -> str:
+        await workflow.wait_condition(lambda: self.state.cluster_started)
+        assert not self.state.cluster_shutdown
```
cretz, line 108: This should probably match the error from allocate (see comment there, this will fail task by default, may prefer `ApplicationError`)

updates_and_signals/atomic_message_handlers/README.md
drewhoskins-temporal: 🤔 I wanted people to see updates and signals for discoverability, and we're planning at least one more updates sample.

cretz: We haven't usually grouped by those top-level features before but more by what the sample does. So we don't have `interceptors/context_propagation` and `interceptors/sentry`, just two top-level separate samples that use the same Temporal features. We just need to determine whether we want this type of grouping now and maybe apply it generally. I know our other samples repositories have also tried to avoid nesting most samples.

updates_and_signals/atomic_message_handlers/starter.py
```diff
@@ -0,0 +1,70 @@
...
+        await asyncio.sleep(delay_seconds)
+
+    deletion_updates = []
+    for i in range(6):
+        deletion_updates.append(
+            wf.execute_update(ClusterManagerWorkflow.delete_job, f"task-{i}")
+        )
+    await asyncio.gather(*deletion_updates)
+
+    await wf.signal(ClusterManagerWorkflow.shutdown_cluster)
```
cretz, line 36: Arguably shutdown could be an update that returns what the workflow returns instead of making it a two-step process (but this is fine too)

drewhoskins-temporal: Cool idea, and would show off the power of update.  Ran out of time this A.M, though.

updates_and_signals/atomic_message_handlers/activities.py
cretz: You can still use named arguments. We use them in lots of places, but since we're the only users of them we don't need to set this setting to force us to use them. Also, we have a CI check for our samples in 3.8 and I don't think this came about until 3.10 (we can look into relaxing our CI version constraints though).

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -0,0 +1,218 @@
...
+    @workflow.signal
+    async def shutdown_cluster(self):
+        await workflow.wait_condition(lambda: self.state.cluster_started)
+        self.state.cluster_shutdown = True
+        workflow.logger.info("Cluster shut down")
+
+    @workflow.update
+    async def allocate_n_nodes_to_job(
+        self, input: ClusterManagerAllocateNNodesToJobInput
+    ) -> List[str]:
```
dandavison, line 83: I think it would be nice to have docstrings on the signals and updates, e.g. explaining what the update returns. I'm thinking that this would help users understand why it's an update and how updates are useful.

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -0,0 +1,218 @@
...
+        await workflow.execute_activity(
+            allocate_nodes_to_job,
+            AllocateNodesToJobInput(nodes=assigned_nodes, task_name=task_name),
+            start_to_close_timeout=timedelta(seconds=10),
+        )
+        for node in assigned_nodes:
+            self.state.nodes[node] = task_name
+
+    @workflow.update
+    async def delete_job(self, input: ClusterManagerDeleteJobInput):
```
dandavison, line 121: This doesn't return anything, so I think readers will be wondering why it's an update rather than a signal.

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -0,0 +1,218 @@
...
+        if self.nodes_lock.locked():
+            return False
+        if workflow.info().is_continue_as_new_suggested():
+            return True
+        # This is just for ease-of-testing.  In production, we trust temporal to tell us when to continue as new.
+        if (
+            self.max_history_length
+            and workflow.info().get_current_history_length() > self.max_history_length
+        ):
+            return True
```
dandavison, line 185: This makes it more confusing from a pedagogical point of view. Might be nice to switch to e.g. using `mock.patch` in the test to control CAN limit. (Non-blocking comment)

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -0,0 +1,218 @@
...
+            unassigned_nodes = [k for k, v in self.state.nodes.items() if v is None]
+            if len(unassigned_nodes) < input.num_nodes:
+                # If you want the client to receive a failure, either add an update validator and throw the
+                # exception from there, or raise an ApplicationError. Other exceptions in the main handler
+                # will cause the workflow to keep retrying and get it stuck.
+                raise ApplicationError(
+                    f"Cannot allocate {input.num_nodes} nodes; have only {len(unassigned_nodes)} available"
+                )
+            assigned_nodes = unassigned_nodes[: input.num_nodes]
+            # This await would be dangerous without nodes_lock because it yields control and allows interleaving.
```
dandavison, line 103: Would help users understand the locking even more if this comment said what it is that shouldn't be interleaved.

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -0,0 +1,218 @@
...
+@dataclass
+class ClusterManagerResult:
+    max_assigned_nodes: int
+    num_currently_assigned_nodes: int
+
+
+@dataclass(kw_only=True)
+class ClusterManagerAllocateNNodesToJobInput:
+    num_nodes: int
+    task_name: str
```
dandavison, line 49: Would be nice to use "job" xor "task" in names.

updates_and_signals/atomic_message_handlers/workflow.py
cretz: Updates that don't return anything are totally reasonable (they can still raise errors for instance and just waiting on their completion means you know it completed, both of which are improvements over signals). However, I would strongly recommend a `-> None` type hint here.

updates_and_signals/atomic_message_handlers/workflow.py
drewhoskins-temporal: Note that it's not just a pytest affordance, it's also for the sample.  (there's a --test-continue-as-new argument)

updates_and_signals/atomic_message_handlers/workflow.py
```diff
@@ -103,25 +106,28 @@ async def allocate_n_nodes_to_job(
...
             allocate_nodes_to_job,
-            AllocateNodesToJobInput(nodes=assigned_nodes, task_name=task_name),
+            AllocateNodesToJobInput(nodes=assigned_nodes, job_name=job_name),
             start_to_close_timeout=timedelta(seconds=10),
         )
         for node in assigned_nodes:
-            self.state.nodes[node] = task_name
+            self.state.nodes[node] = job_name
 
+    # Even though it returns nothing, this is an update because the client may want track it, for example
```
dandavison, line 127:
```suggestion
    # Even though it returns nothing, this is an update because the client may want to track it, for example
```

README.md
```diff
@@ -52,6 +52,7 @@ Some examples require extra dependencies. See each sample's directory for specif
   * [hello_signal](hello/hello_signal.py) - Send signals to a workflow.
 <!-- Keep this list in alphabetical order -->
 * [activity_worker](activity_worker) - Use Python activities from a workflow in another language.
+* [atomic_message_handlers](updates_and_signals/atomic_message_handlers/) - Safely handling updates and signals.
```
dandavison, line 55: I think the name of the sample should be changed to something like `safe_message_handling`. It's not about atomicity -- the sample doesn't demonstrate rolling back of incomplete side effects. Rather it's about maintaining strict isolation between handler executions, via serialization of handler executions. In any case, we don't want users to think this is showing a specialized form of message handling that they can ingnore; we want them to consider whether they need this for any workflow with message handlers.

drewhoskins-temporal: Good idea.  "Safe" feels much more like something I'm supposed to read.

cretz: Not the biggest fan of "safe" vs "atomic" since the latter is more discoverable/descriptive when looking at the list of samples, but I don't have a strong opinion here.

dandavison: @cretz we could choose a word other than "safe", but I argued above that "atomic" isn't the right word.

cretz: I don't think "atomic" relates to rollback at all. Atomic just means one at a time, as opposed to "transactional". I think atomic is a good word, but again I don't have a strong opinion.

dandavison: Yes, to be honest I'm not in love with "safe" and implying that any other usage style is not safe.

Hm, an atomic operation is one that either completes in its entirety or behaves as if it never started, and can't be seen in an intermediate state. So, if the operation has multiple stages with side effects, that would require some notion of rollback. It's usually synonymous with "transactional". I agree it's closely related to the idea of serializing executions so that they occur one at a time, since that's one way of ensuring that one execution can't see in-progress state of another, but using "atomic" would imply that message handling that does multiple writes can rollback incomplete changes. I think here we're talking about "serialized message processing" or "preventing corruption of shared state by message handlers".

drewhoskins-temporal: There's more here than just concurrency, such as dangling handlers.  Sticking with safe.
I'm think I'm going to touch on idempotency as well in my next push, though we probably should also add a more focused idempotency sample.

updates_and_signals/atomic_message_handlers/activities.py
drewhoskins-temporal: Ah, too bad.  I'd rather people who pattern-match off of this sample be directed toward best practices.  Will remove for now.  I wonder if we have stats on python versions people actually use in the wild?

README.md
drewhoskins-temporal: > Yes, to be honest I'm not in love with "safe" and implying that any other usage style is not safe.

I don't think it implies that.  "Robust" is an alternate word.

drewhoskins-temporal: Update: added idempotency.  I didn't use the built-in update ID, though I could

README.md
```diff
@@ -52,6 +52,7 @@ Some examples require extra dependencies. See each sample's directory for specif
   * [hello_signal](hello/hello_signal.py) - Send signals to a workflow.
 <!-- Keep this list in alphabetical order -->
 * [activity_worker](activity_worker) - Use Python activities from a workflow in another language.
+* [safe_message_handlers](updates_and_signals/safe_message_handlers/) - Safely handling updates and signals.
```
cretz, line 55: Should keep this list in alphabetical order (see comment a couple of lines above). Also not a fan of nesting these non-hello samples beneath a directory unnecessarily (you'll note we don't do this much in other samples here or in many samples repos). If you _must_ inconsistently nest this sample, you may want nested bullets here.

tests/updates_and_signals/safe_message_handlers_test.py
cretz, line 1: We prefer tests to be in the same directory under `tests` that they are in the top level. So `/custom_converter/` tests are in `/tests/custom_converter/` and therefore `/updates_and_signals/safe_message_handlers/` tests should be under `/tests/updates_and_signals/safe_message_handlers/` (granted as mentioned in comments before, I don't think we should nest sample dirs).

updates_and_signals/safe_message_handlers/README.md
```diff
@@ -0,0 +1,21 @@
...
+* You can run start_workflow with an initializer signal that you want to run before anything else other than the workflow's constructor.  This pattern is known as "signal-with-start."
+* Message handlers can block and their actions can be interleaved with one another and with the main workflow.  This can easily cause bugs, so we use a lock to protect shared state from interleaved access.
+* Message handlers should also finish before the workflow run completes.  One option is to use a lock.
+* An "Entity" workflow, i.e. a long-lived workflow, periodically "continues as new".  It must do this to prevent its history from growing too large, and it passes its state to the next workflow.  You can check `workflow.info().is_continue_as_new_suggested()` to see when it's time.  Just make sure message handlers have finished before doing so.  
+* Message handlers can be made idempotent.  See update `ClusterManager.allocate_n_nodes_to_job`.
+
+To run, first see [README.md](../../README.md) for prerequisites.
+
+Then, run the following from this directory to run the sample:
```
cretz, line 14: Sometimes people get confused that they can't just run these two commands in the same terminal because the first blocks. In our sample READMEs we usually make clear that the starter needs to be in a separate terminal.

updates_and_signals/safe_message_handlers/activities.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@dataclass
+class AllocateNodesToJobInput:
+    nodes: List[str]
+    job_name: str
+
+
+@activity.defn
+async def allocate_nodes_to_job(input: AllocateNodesToJobInput):
```
cretz, line 15: Should provide type hints for every activity return, even if `-> None`, it helps callers

updates_and_signals/safe_message_handlers/starter.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+async def main(should_test_continue_as_new: bool):
+    # Connect to Temporal
+    client = await Client.connect("localhost:7233")
+
+    cluster_manager_handle = await client.start_workflow(
+        ClusterManagerWorkflow.run,
+        ClusterManagerInput(test_continue_as_new=should_test_continue_as_new),
+        id=f"ClusterManagerWorkflow-{uuid.uuid4()}",
```
cretz, line 57: In other samples we have used fixed workflow IDs, but don't technically have to here, but it makes the `id_reuse_policy` have no value since this is always unique

updates_and_signals/safe_message_handlers/starter.py
```diff
@@ -0,0 +1,80 @@
...
+        task_queue="atomic-message-handlers-task-queue",
```
cretz, line 58: Task queue was not changed to match the sample name

updates_and_signals/safe_message_handlers/worker.py
```diff
@@ -0,0 +1,41 @@
...
+    client = await Client.connect("localhost:7233")
+
+    async with Worker(
+        client,
+        task_queue="atomic-message-handlers-task-queue",
+        workflows=[ClusterManagerWorkflow],
+        activities=[allocate_nodes_to_job, deallocate_nodes_for_job, find_bad_nodes],
+    ):
+        # Wait until interrupted
+        logging.info("ClusterManagerWorkflow worker started, ctrl+c to exit")
```
cretz, line 29: Inconsistent use of logging vs print

cretz (commented):
Some minor things. Also there seems to be a test failure in CI.

updates_and_signals/safe_message_handlers/README.md
drewhoskins-temporal: I copy-pasted this.  Looks like I got unlucky in which one I chose!

cretz: :+1: Yeah we are admittedly not consistent, this is not a blocker or anything.

cretz (approved):

dandavison (approved):
This is great. Feel free to leave my comments for follow-on PRs (and flag those you disagree with on any timescale).

cretz:
CI needs to pass then we can merge

drewhoskins-temporal:
This is blocked on the Java Test Service being fixed for updates (or switching those tests away from using that time-skipping service)

PR #125, opened 2024-07-08
Refactor: Correct polling infrequent/periodic_sequence usecase

pfranck:
## What was changed
We use the activity.info().attempt value to feed the TestService

## Why?
Before this change the attempt value in the TestService was never change and stay to 0

## Checklist
2. How was this tested:
Run the 2 command explain in the readme - success on 5 atempts

### DIFF
--- a/polling/infrequent/activities.py
+++ b/polling/infrequent/activities.py
@@ -11,10 +11,12 @@ class ComposeGreetingInput:
     name: str
 
 
-@activity.defn
-async def compose_greeting(input: ComposeGreetingInput) -> str:
-    attempt = activity.info().attempt - 1
-    test_service = TestService(attempt=attempt)
-    # If this raises an exception because it's not done yet, the activity will
-    # continually be scheduled for retry
-    return await test_service.get_service_result(input)
+class ComposeGreeting:
+    def __init__(self):
+        self.test_service = TestService()
+
+    @activity.defn
+    async def compose_greeting(self, input: ComposeGreetingInput) -> str:
+        # If this raises an exception because it's not done yet, the activity will
+        # continually be scheduled for retry
+        return await self.test_service.get_service_result(input)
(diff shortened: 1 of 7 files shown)


polling/infrequent/activities.py
```diff
@@ -13,7 +13,8 @@ class ComposeGreetingInput:
 
 @activity.defn
 async def compose_greeting(input: ComposeGreetingInput) -> str:
-    test_service = TestService()
+    attempt = activity.info().attempt - 1
+    test_service = TestService(attempt=attempt)
```
cretz, line 17: Thanks! Seems this was written with a bug.

I think a cleaner approach would be to make this activity a method (and call via `execute_activity_method` from workflow) and instantiate the `TestService` as class state.

Also, I think  `frequent` polling is affected by this same test service issue.

polling/infrequent/activities.py
```diff
@@ -10,10 +10,12 @@ class ComposeGreetingInput:
     greeting: str
     name: str
 
+class ComposeGreeting:
+    def __init__(self, test_service: TestService = None):
```
cretz, line 14: I think this has to be `Optional` to accept a default of `None`. But probably don't even need parameter, just instantiate in constructor always.

polling/test_service.py
```diff
@@ -1,7 +1,7 @@
 class TestService:
-    def __init__(self):
+    def __init__(self, error_attempts=5):
```
cretz, line 2: Unused parameter, can probably just remove

cretz (commented):
Looks good, thanks! May need reformat (`poe format`) and may need to check types (`poe lint-types`). If you want you can also apply to the other polling test that uses this test service, but no worries if not, we can create another issues.

pfranck:
`poe format` and `poe lint-types` have been runned

brianstrauch:
The underlying issue has since been fixed in main via #152, which reworked `polling/test_service.py` and the `infrequent`/`periodic_sequence` samples (plus `frequent`). The attempt counter now increments correctly per workflow, and `periodic_sequence` matches its README.

Closing as superseded — appreciate you catching the bug.

PR #131, opened 2024-07-23
Updates for temporal CLI over tctl and temporal dev-server

cici:
## What was changed
I updated the README to use the ```temporal``` CLI instead of ```tctl``` and updated codec_server.py to default to port 8233 for the local web UI since that is the default port for ```temporal server start-dev```

## Checklist

1. Closes 

2. How was this tested:

Tested on my local machine

3. Any docs updates needed?

Done

### DIFF
--- a/encryption/README.md
+++ b/encryption/README.md
@@ -4,10 +4,6 @@ This sample shows how to make an encryption codec for end-to-end encryption. It 
 samples [in TypeScript](https://github.com/temporalio/samples-typescript/tree/main/encryption) and
 [in Go](https://github.com/temporalio/samples-go/tree/main/encryption).
 
-Ensure you have an environment variable set to the Namespace your Workflows are in:
-
-
-    export TEMPORAL_NAMESPACE=<Namespace Name>
 
 For this sample, the optional `encryption` dependency group must be included. To include, run:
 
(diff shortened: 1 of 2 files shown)


encryption/README.md
```diff
@@ -4,6 +4,11 @@ This sample shows how to make an encryption codec for end-to-end encryption. It
 samples [in TypeScript](https://github.com/temporalio/samples-typescript/tree/main/encryption) and
 [in Go](https://github.com/temporalio/samples-go/tree/main/encryption).
 
+Ensure you have an environment variable set to the Namespace your Workflows are in:
+
+
+    export TEMPORAL_NAMESPACE=<Namespace Name>
+
```
cretz, line 11: This variable is not used in this sample, the namespace is fixed to `default` (we expect to come later and make all samples run somewhere configurable)

cretz (commented):
Thanks! One minor note about the env var.

encryption/README.md
cici: I added that in because my TEMPORAL_NAMESPACE was set to my Cloud Namespace and when I ran the starter, I got an error that the Namespace didn't exist, so I reset it to "default" and then it ran fine.

cretz: The starter in the encryption sample doesn't use env vars. Do you maybe mean the `temporal` CLI which does? There may be other env vars the CLI uses that you would have to make sure are unset if you're concerned about your existing environment variable leftovers polluting the CLI (though I can't think of any for dev server). Also, would encourage `unset TEMPORAL_NAMESPACE` instead of `<Namespace Name>` and "Namespace your Workflows are in". This sample is hardcoded to work with `default`.

cici: I wasn't using the CLI. I do see where it is using default, so I will have to debug what I was seeing. I will remove that statement.

cretz (approved):
Looks great! (let me know if it doesn't let you click the merge button and I'll click it)

PR #132, opened 2024-07-24
Remove health check and "bad" node concept from ClusterManager

dandavison:
A concern was raised that our samples should not be demonstrating any form of polling from within an entity workflow loop, even if the poll frequency is low. Instead, we intend to point to long-running activity patterns.


updates_and_signals/safe_message_handlers/workflow.py
```diff
@@ -229,9 +196,7 @@ def should_continue_as_new(self) -> bool:
     async def run(self, input: ClusterManagerInput) -> ClusterManagerResult:
         self.init(input)
         await workflow.wait_condition(lambda: self.state.cluster_started)
-        # Perform health checks at intervals.
         while True:
```
cretz, line 199: There is no value in this being a loop anymore and no value in the wait condition having a timeout (and therefore no value in a sleep interval setting)

dandavison: Ah, very good point! (Fixed same bug in Typescript version). I think this actually improves the sample a lot, since it may not be obvious to users that they can implement the main "loop" so simply.

updates_and_signals/safe_message_handlers/workflow.py
```diff
@@ -228,29 +193,16 @@ def should_continue_as_new(self) -> bool:
...
-                        test_continue_as_new=input.test_continue_as_new,
-                    )
-                )
-        return ClusterManagerResult(
-            len(self.get_assigned_nodes()),
-            len(self.get_bad_nodes()),
+        await workflow.wait_condition(
+            lambda: self.state.cluster_shutdown or self.should_continue_as_new()
         )
+        if self.should_continue_as_new():
```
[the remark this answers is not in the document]
dandavison: Right, I'm trying to move too quickly here. Reinstated loop, and added wait for `all_handlers_finished()` before CAN. It would be ideal to add some tests exercising CAN.

cretz (approved):

dandavison:
This is now blocked on an `sdk-python` release.

drewhoskins-temporal (changes requested):
The purpose is to demonstrate how to use a mutex to protect interleaving between the handlers and the main coroutine and to show something that solves a real-world problem where events come both from the workflow itself and from outside.  Is there an alternate way to code up a health check that wouldn't involve polling?

PR #138, opened 2024-08-28
Adding encryption example using a KMS and JWT-based auth

phillipskevin:


encryption_jwt/codec_server.py
```diff
@@ -0,0 +1,119 @@
...
+        async def handler(req: web.Request):
+            # Read payloads as JSON
+            assert req.content_type == "application/json"
+            payloads = json_format.Parse(await req.read(), Payloads())
+
+            # Extract the email from the JWT.
+            auth_header = req.headers.get("Authorization")
+            namespace = req.headers.get("x-namespace")
+            _bearer, encoded = auth_header.split(" ")
+            decoded = jwt.decode(encoded, options={"verify_signature": False})
```
cretz, line 78: Are we sure we don't want to verify this signature against the cloud's public key? Ref https://docs.temporal.io/production-deployment/data-encryption#authorization, we probably want to check against https://login.tmprl.cloud/.well-known/jwks.json or it's subject to MITM.

cretz:
Thanks! Added some review comments. Also may want to `poe format` and `poe lint-types`.

encryption_jwt/codec_server.py
```diff
@@ -0,0 +1,119 @@
...
+        resp.headers[hdrs.ACCESS_CONTROL_ALLOW_HEADERS] = allow_headers
+
+        return resp
+
+    def decryption_authorized(email: str, namespace: str) -> bool:
+        credentials = grpc.composite_channel_credentials(grpc.ssl_channel_credentials(
+        ), grpc.access_token_call_credentials(os.environ.get("TEMPORAL_API_KEY")))
+
+        with grpc.secure_channel(temporal_ops_address, credentials) as channel:
+            client = service_pb2_grpc.CloudServiceStub(channel)
```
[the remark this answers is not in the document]
phillipskevin: @cretz is there an example of using this? I was trying trying to get it to work but running into errors and some of the types aren't documented yet, so it was tough to figure out what I'm doing wrong.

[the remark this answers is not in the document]
cretz: No, it is still experimental and this may not be its final form (and even the gRPC service could change, it is not GA yet). It's basically direct gRPC similar to what you're doing. Here is a test case where we call it: https://github.com/temporalio/sdk-python/blob/2e6d742ce1531741c0f953a311689bd72781b720/tests/test_client.py#L1315-L1325.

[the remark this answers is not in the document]
phillipskevin: Thanks! I got this working 👍

phillipskevin:
Thanks for all the suggestions, @cretz. I think I've addressed everything. Let me know if you have any more feedback.

encryption_jwt/encryptor.py
```diff
@@ -0,0 +1,83 @@
...
+        data_key_encrypted, data_key_plaintext = await self.__create_data_key(
+            self._namespace
+        )
+
+        if data_key_encrypted is None:
+            raise ValueError("No data key!")
+
+        nonce = os.urandom(12)
+        encryptor = AESGCM(data_key_plaintext)
+        return nonce + encryptor.encrypt(nonce, data, None), base64.b64encode(
```
cretz, line 41:
```suggestion
        encrypted = asyncio.get_running_loop().run_in_executor(None, encryptor.encrypt, nonce, data, None)
        return nonce + encrypted, base64.b64encode(
```

I am a little worried about using this under load with CPU bound encryption (and decryption below) blocking the event loop. Technically what is here is ok, but this may make it better, thoughts? Or at least a comment above this encrypt line and the decrypt equivalent below mentioning it? (technically could have put the entire boto3 thread-blocking call under a new thread with this encrypt too if you wanted, sorry I didn't mention it before instead of using aioboto3)

cretz:
Make need to run `poe lint` to confirm type checking.

phillipskevin:
Should be good to go now, @cretz. Thanks again!

cretz:
@phillipskevin - Looks like there's a type error. Specifically, since we allow our Python samples to run in in all non-EOL Python versions, you have to change things like `tuple[bytes, bytes]` to `typing.Tuple[bytes, bytes]`.

For this error:

> Library stubs not installed for "requests" (or incompatible with Python 3.8)

You may need to add a dependency for https://pypi.org/project/types-requests. Have not investigated this.

pyproject.toml
```diff
@@ -43,6 +43,10 @@ dependencies = { pyyaml = "^6.0.1", types-pyyaml = "^6.0.12", dacite = "^1.8.1"
 optional = true
 dependencies = { cryptography = "^38.0.1", aiohttp = "^3.8.1" }
 
+[tool.poetry.group.encryption_jwt]
+optional = true
+dependencies = { cryptography = "^38.0.1", aiohttp = "^3.8.1", pyjwt = "^2.9.0", aioboto3 = "^13.1.1", "requests" = "^2.32.3", "types-requests": "^2.32.0" }
```
cretz, line 48:
```suggestion
dependencies = { cryptography = "^38.0.1", aiohttp = "^3.8.1", pyjwt = "^2.9.0", aioboto3 = "^13.1.1", "requests" = "^2.32.3", "types-requests" = "^2.32.0" }
```

phillipskevin:
Looks like CI is still failing after adding the `types-requests` dependency. I'll try to look into why.

cretz:
Looks like it's still complaining on 3.8. I admittedly have not spent time digging into this.

phillipskevin:
Python 3.8 went end of life last week, so having some trouble testing it now. Will the versions used in CI be updated? Or will you continue to support 3.8 for some time?

cretz:
I suspect we will EOL it too soon, though no specific timeline

cretz:
Opened https://github.com/temporalio/sdk-python/issues/672

dlee-bitovi:
Hey @cretz , Happy Friday!

Here's some screenshots of my recent commits to this PR that successfully run the below commands locally - would you be willing to approve the workflows to run (when you get a chance)?
1. `uv sync`
2. `poe lint`
3. `poe test`

Thanks!
image

cretz:
Approved and running CI now, thanks for sticking with this! Also, if possible, any kind of tests to confirm this remains working would be helpful (we have had some contributed samples become stale and stop working for one reason or another).

Mattchewone (approved):

encryption_jwt/codec_server.py
```diff
@@ -0,0 +1,165 @@
...
+        return handler
+
+    # Build app
+    app = web.Application()
+    # set up logger
+    logging.basicConfig(level=logging.DEBUG)
+    logger = logging.getLogger(__name__)
+    app.add_routes(
+        [
+            web.post("/encode", make_handler("encode")),
```
Mattchewone, line 144: Do we need to expose the encode over http?

cretz (commented):
If possible, any tests would help here, though I know it can be a bit difficult since the cloud JWKS is hardcoded. We can discuss merging without tests if they are too hard to write or they need to be done after merge. (we know many of our samples like the encryption sample don't have tests and we keep getting burned by not having them)

PR #139, opened 2024-08-31
Demonstrate workflow.all_handlers_finished

drewhoskins-temporal:
## What was changed

Before `workflow.all_handlers_finished` existed, I worked around it by checking a lock to see if we were in the middle of any handlers before continuing as new.
Now in 1.7, we can use the new method to prevent dangling handlers.

## Why?

Demonstrate what should become standard practice.

## Checklist

1. Closes 

2. How was this tested:

```
# Terminal 1
cd updates_and_signals/safe_message_handlers
poetry run python ./worker.py
# Terminal 2
cd updates_and_signals/safe_message_handlers
poetry run python starter.py --test-continue-as-new
```
3. Any docs updates needed?

Yes, see the README.md


dandavison (approved):
LGTM

PR #140, opened 2024-09-07
feat: add example using Sentry V2 SDK

gregbrowndev:
## What was changed

Added a new example to set up Sentry using the V2 SDK.

## Why?

After recently trying to set up Sentry, the existing example that uses Sentry's V1 SDK doesn't work for me. The issue seems to be related to the `warnings` and `threading` modules that Sentry uses not being appropriately included in the Temporal Workflow's sandbox. This results in the Workflow execution failing to start in the worker that handles the Workflow. I don't see any issues with the Sentry interceptor on other workers that only handle activities, so it seems the problem is related to the Workflow sandbox.

Note: I also tried disabling the worker's sandbox, which seemed to get further. The workflow is able to start and eventually finish, but you get a sporadic error in the workflow's event history.

The solution is to migrate the interceptor to use V2 of Sentry's SDK, where the Hub object is now deprecated in favour of using scopes. Using the scope context manager doesn't seem to have the same issues with the warnings and threading libs (which it still uses).

I've created a new example to keep the V1 example in place, since V2 only works with Python 3.6 or above.

## Checklist

1. Closes 

https://temporalio.slack.com/archives/CTT84RS0P/p1725394300914329

2. How was this tested:

Tested manually in my application using Sentry.io

3. Any docs updates needed?

Added a new example with a README explaining the reasons/difference between the original example and the v2 example.

### DIFF
--- a/sentry_v2/README.md
+++ b/sentry_v2/README.md
@@ -1,0 +1,33 @@
+# Sentry V2 Sample
+
+This sample shows how to configure [Sentry](https://sentry.io) SDK V2 to intercept and capture errors from the Temporal SDK.
+
+Note: This is a small modification of the original example [sentry (v1)](../sentry) which seems to have
+issues in the Workflow interceptor, as there are modules (`warnings` and `threading`) that Sentry uses that aren't passed through
+to Temporal's sandbox properly, causing the Workflows to fail to execute.
+
+> Sentry V2 only supports Python 3.6 and higher. So if you're on Python 3.5, you may need to refer to the original example
+> (which doesn't work for me) or upgrade your Python version.
+
+### Further details
+
+Sentry's `Hub` object is now deprecated in the V2 SDK in favour of scopes. See [Activating Current Hub Clone](https://docs.sentry.io/platforms/python/migration/1.x-to-2.x#activating-current-hub-clone)
+for more details. The changes are  simple, just replace `with Hub(Hub.current):` with `with isolation_scope() as scope:`.
+
+## Running the Sample
+
+For this sample, the optional `sentry` dependency group must be included. To include, run:
+
+    poetry install --with sentry
+
+To run, first see [README.md](../README.md) for prerequisites. Set `SENTRY_DSN` environment variable to the Sentry DSN.
+Then, run the following from this directory to start the worker:
+
+    poetry run python worker.py
+
+This will start the worker. Then, in another terminal, run the following to execute the workflow:
+
+    poetry run python starter.py
+
+The workflow should complete with the hello result. If you alter the workflow or the activity to raise an
+`ApplicationError` instead, it should appear in Sentry.
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .venv
 __pycache__
+.idea
--- a/poetry.lock
+++ b/poetry.lock
@@ -2128,21 +2128,22 @@ crt = ["botocore[crt] (>=1.33.2,<2.0a.0)"]
 
 [[package]]
 name = "sentry-sdk"
-version = "1.45.0"
+version = "2.13.0"
 description = "Python client for Sentry (https://sentry.io)"
 optional = false
-python-versions = "*"
+python-versions = ">=3.6"
 files = [
-    {file = "sentry-sdk-1.45.0.tar.gz", hash = "sha256:509aa9678c0512344ca886281766c2e538682f8acfa50fd8d405f8c417ad0625"},
-    {file = "sentry_sdk-1.45.0-py2.py3-none-any.whl", hash = "sha256:1ce29e30240cc289a027011103a8c83885b15ef2f316a60bcc7c5300afa144f1"},
+    {file = "sentry_sdk-2.13.0-py2.py3-none-any.whl", hash = "sha256:6beede8fc2ab4043da7f69d95534e320944690680dd9a963178a49de71d726c6"},
+    {file = "sentry_sdk-2.13.0.tar.gz", hash = "sha256:8d4a576f7a98eb2fdb40e13106e41f330e5c79d72a68be1316e7852cf4995260"},
 ]
 
 [package.dependencies]
 certifi = "*"
-urllib3 = {version = ">=1.26.11", markers = "python_version >= \"3.6\""}
+urllib3 = ">=1.26.11"
 
 [package.extras]
 aiohttp = ["aiohttp (>=3.5)"]
+anthropic = ["anthropic (>=0.16)"]
 arq = ["arq (>=0.23)"]
 asyncpg = ["asyncpg (>=0.23)"]
 beam = ["apache-beam (>=2.12)"]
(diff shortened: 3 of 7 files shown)


sentry_v2/README.md
```diff
@@ -0,0 +1,33 @@
+# Sentry V2 Sample
+
+This sample shows how to configure [Sentry](https://sentry.io) SDK V2 to intercept and capture errors from the Temporal SDK.
+
+Note: This is a small modification of the original example [sentry (v1)](../sentry) which seems to have
+issues in the Workflow interceptor, as there are modules (`warnings` and `threading`) that Sentry uses that aren't passed through
+to Temporal's sandbox properly, causing the Workflows to fail to execute.
```
cretz, line 7: If Sentry V1 doesn't work anymore, let's replace the sample instead of adding a new one

sentry_v2/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+> Sentry V2 only supports Python 3.6 and higher. So if you're on Python 3.5, you may need to refer to the original example
```
cretz, line 9: Temporal Python SDK only supports 3.8 or higher, so this disclaimer is not needed

sentry_v2/README.md
gregbrowndev: Hey @cretz - I haven't been able to test this, but I suspect now the example didn't work for me because when I installed sentry-sdk, it will have defaulted to the latest version (V2), but the example only works with V1. I imagine if you pin the version to the one in the example, it might work.

It does look like development on the V1 SDK has stopped now (last feature release was [in April](https://github.com/getsentry/sentry-python/releases/tag/1.45.0) with a security patch [in July](https://github.com/getsentry/sentry-python/releases/tag/1.45.1)). 

So, I will let you decide if we should replace the V1 example. I'll update the PR as necessary.

cretz: Hrmm, assuming the current `sentry` works, how about we just move the current one to `sentry_v1` and make yours just `sentry`? Then we can come back and delete `sentry_v1` later when they officially EOL it.

gregbrowndev: @cretz I renamed sentry -> sentry_v1 and kept sentry_v2 as is, so the commit history doesn't get skewed for the original impl. Let me know if you'd prefer to make sentry_v2 -> sentry.

cretz: > Let me know if you'd prefer to make sentry_v2 -> sentry.

That does seem to make more sense, but I don't know Sentry's stance on supporting v1. If we support v1 and v2, I think we should have different groups.

cretz: Does Sentry still support v1? If it is EOL, maybe we should consider it EOL too.

sentry_v2/README.md
cretz: > you cannot have two different poetry groups containing the same package with different versions

Ug, that is unfortunate. I do still think we should have the Poetry groups named after the sample they represent, even if it's a dupe for `sentry_v1` and `sentry_v2` (or there isn't even a `sentry_v1` group at all and pip install is manual). This also may give more credence to the approach of making `sentry_v2` just `sentry`.

gregbrowndev: I've [asked on the Sentry's discussion board](https://github.com/getsentry/sentry-python/discussions/2980#discussioncomment-11291164) what the timeline is for v1 support. Will get back once they provide an answer.

One idea would be to just replace the v1 impl. with the new v2 integration, keep the example called `sentry`, and then add a link in the README to the original v1 commit for posterity?

cretz: Sure, that would work well. Either way I think we want this to be the primary `sentry` sample since we have to have the dependency group named that.

cretz:
Uh oh, v3? Ok, I guess we will cross that bridge when we get there. Otherwise, I think this looks good. And the purpose is to serve as a sample, so users can adapt it as they need for their version of tooling.

gregbrowndev:
Hey @cretz, do you want me to rebase the PR?

cretz:
@gregbrowndev - I can, no problem. Looks like just the CLA needs to be signed (see link above pointing to https://cla-assistant.io/temporalio/samples-python?pullRequest=140).

gregbrowndev:
Sorry @cretz, I've only just seen that this PR was blocked by me not signing the CLA. I've done so now. 

Could you comment on @M0NsTeRRR 's message above? Is the `SandboxedWorkflowRunner` passthrough restrictions the same thing as using the decorator in the workflow like below or are both necessary? I haven't looked at the project I used the Sentry integration in a while, but can re-test this sample again if I get some time today/tomorrow

```py
    import sentry_sdk

    from sentry.interceptor import SentryInterceptor
```

M0NsTeRRR:
After reading it again, I realized that I didn’t import the interceptor in my workflow, which is likely why I got this error — my bad. However, I think it makes more sense for the example to whitelist it at the worker level, since every exception will be caught by the Sentry interceptor.

By the way, not sure if you forgot, but the PR hasn’t been merged because you haven’t signed the CLA yet, @gregbrowndev 🙂.
https://github.com/temporalio/samples-python/pull/140#issuecomment-2511566179

cretz:
> I've only just seen that this PR was blocked by me not signing the CLA. I've done so now.

Thanks! Would you be willing to merge `main` and resolve conflicts?

> Is the SandboxedWorkflowRunner passthrough restrictions the same thing as using the decorator in the workflow like below or are both necessary?

Yes, they are the same thing, but note some imports are done lazily by some libraries so we have to account for that too. What we usually do in newer samples and what we encourage users to do (and would like to here, though don't have to now), is to have workflows be in their own file instead of a shared file.

sentry/interceptor.py
```diff
@@ -12,59 +12,63 @@
...
                 if len(input.args) == 1 and is_dataclass(input.args[0]):
-                    set_context("temporal.workflow.input", asdict(input.args[0]))
-                set_context("temporal.workflow.info", workflow.info().__dict__)
+                    scope.set_context("temporal.workflow.input", asdict(input.args[0]))
+                scope.set_context("temporal.workflow.info", workflow.info().__dict__)
 
                 if not workflow.unsafe.is_replaying():
                     with workflow.unsafe.sandbox_unrestricted():
-                        capture_exception()
+                        scope.capture_exception()
```
Natim, line 71: :heart:

Natim:
```suggestion
                        try:
                            scope.capture_exception()
                        except TypeError:
                            logger.exception("An error occured when trying to call Sentry from the interceptor.")
```

sentry/worker.py
```diff
@@ -1,64 +1,92 @@
...
         client,
         task_queue="sentry-task-queue",
         workflows=[GreetingWorkflow],
         activities=[compose_greeting],
         interceptors=[SentryInterceptor()],  # Use SentryInterceptor for error reporting
-    )
-
-    await worker.run()
+        workflow_runner=SandboxedWorkflowRunner(
+            restrictions=SandboxRestrictions.default.with_passthrough_modules(
```
Natim, line 65: Should it be:

```suggestion
            restrictions=SandboxRestrictions.default().with_passthrough_modules(
```

cretz:
> To confirm, I saw the "metaclass conflict" error [...]

Is it possible to add some test that doesn't require Sentry for this sample? Our Sentry V1 sample started breaking and we didn't know because there were no tests. Also, when the test is added, we can easily replicate this error and I can help debug it.

> As already mentioned, I found an issue with uv. Because the gevent sample is included in default-groups, it installs the gevent library, which breaks Sentry's isolation_scope as it relies on setting contextvars.

Yeah, we have an issue to remove default groups, they were added by accident: #190 

>Lastly, I noticed that the Worker SDK is using a _ShutdownRequested exception to trigger shutdowns. I guess because the exception is never extracted from wait_task = asyncio.wait(tasks, return_when=asyncio.FIRST_EXCEPTION) the AsyncioIntegration` raises it as an unhandled error.

The worker just uses this inside its function to interrupt the `wait`, it is never raised outside of that call (I don't believe wait raises in this instance)

gregbrowndev:
@cretz 

> Is it possible to add some test that doesn't require Sentry for this sample? 

Yes, good idea I will add these tests. I should get a bit of time this weekend

> Yeah, we have an issue to remove default groups, they were added by accident: https://github.com/temporalio/samples-python/issues/190

Sounds good 👍🏻 

> The worker just uses this inside its function to interrupt the wait, it is never raised outside of that call (I don't believe wait raises in this instance)

It's not that the exception bubbles out of the worker, it's just that the `wait` doesn't extract the result of the coroutine explicitly, so Sentry treats it as an unhandled exception, similar to if you did `asyncio.create_task` but didn't read the result. It's no big deal, I just wanted to raise it.

gregbrowndev:
@cretz one last thing to mention. I've been working on a fix for the `TracingInterceptor` in the Python SDK. It doesn't propagate the trace context for process pool workers so log-correlation and other tracing features don't work within our activity impl. inside the subprocess, not sure if its been fixed yet. I've got a working implementation, just trying to get the tests to pass.

I noticed it was easy to break the Sentry interceptor's reflection features, e.g. `input.fn.__module__`, the way I did it initially, and because the Sentry interceptor is in user land, I didn't notice until I deployed onto staging.

Would you consider making the Sentry interceptor a native integration of the Python SDK? I'd be happy to open a PR to the main SDK repo! I'm also working on a Sentry interceptor for the TypeScript SDK too which doesn't have any user-provided samples yet.

sentry/interceptor.py
```diff
@@ -12,63 +13,65 @@
...
             except Exception as e:
                 if len(input.args) == 1:
                     [arg] = input.args
                     if is_dataclass(arg) and not isinstance(arg, type):
-                        set_context("temporal.activity.input", asdict(arg))
-                set_context("temporal.activity.info", activity.info().__dict__)
-                capture_exception()
+                        scope.set_context("temporal.activity.input", asdict(arg))
+                scope.set_context("temporal.activity.info", activity.info().__dict__)
+                scope.capture_exception()
```
kenzmed, line 46: It didn't capture anything for me until I've changed this to:
`sentry_sdk.capture_exception()`

gregbrowndev:
@cretz - merged main and tested/linted locally on 3.10 through 3.13, so hopefully CI will pass 👍🏻

PR #143, opened 2024-10-02
Use workflow init from samples

drewhoskins-temporal:
## What was changed

Workflow Init is a brand new feature.
Let's show people best practices for each sample that uses a signal or update handler and receives workflow arguments.

## Why?

These changes make sure these workflows are initialized before signal or update handlers are run.

## Checklist

1. Closes 

2. How was this tested:

Automated:
`poetry run pytest workflow_test.py`
Manual:
Follow instructions in
`bedrock/entity/README.md`
and
`message_passing/safe_message_handlers/README.md`

3. Any docs updates needed?


pyproject.toml
```diff
@@ -17,7 +17,7 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.8"
-temporalio = "^1.7.0"
+temporalio = { path = "../sdk-python" }
```
drewhoskins-temporal, line 20: Cannot be committed now.  Needs to await the released Python SDK having Workflow Init support.

cretz (commented):
LGTM, holding approval until tagged release of SDK is made

brianstrauch:
Status update: I've brought this branch up to date with `main` (merge, no force-push, so the original commits are intact).

Most of this PR landed through other paths while it sat, so the diff is now a single file.

**Already on `main`:**
- `message_passing/safe_message_handlers/workflow.py` — #153 ("Use workflow.init, refactor") made the same change in Dec 2024: same decorator, same `def init(self, input)` method removed, same lines moved into the constructor.
- `message_passing/introduction/workflows.py` — the comment this PR fixed (`See GreetingWorkflowWithAsyncHandler below`) was deleted entirely by #148, so there's nothing left to correct.

**Dropped:**
- `poetry.lock` — deleted from the repo in #170 (migrate to uv).
- `pyproject.toml` — this was only the temporary `temporalio = { path = "../sdk-python" }` local-path dep, which was never meant to merge.

**Remaining:** `@workflow.init` on `EntityBedrockWorkflow` (`bedrock/entity/workflows.py`, +10/-6). It takes workflow arguments and has signal handlers that mutate the same state those arguments initialize, so it's a good fit for the pattern.

Verified with ruff, mypy, and an end-to-end run against the time-skipping test environment with a stubbed Bedrock activity, sending a signal alongside a pre-populated `prompt_queue` to confirm the constructor's params are applied before the signal is handled. Note there are no automated tests covering the bedrock samples, so that check was manual.

Note the PR description above is now stale — it describes the broader multi-sample scope and references `poetry run pytest`.

PR #144, opened 2024-10-02
Waiting for handlers to finish in all exit cases + abort and compensation in a message handler

dandavison:
Two new samples. From the READMEs:

```markdown
# Waiting for message handlers

This workflow demonstrates how to wait for signal and update handlers to
finish in the following circumstances:

- Before a successful return
- On failure
- On cancellation
```

```markdown
# Waiting for message handlers, and performing compensation and cleanup in message handlers

This sample demonstrates how to do the following:

1. Ensure that all update/signal handlers are finished before a successful
    workflow return, and on workflow cancellation and failure.
2. Perform compensation/cleanup in an update handler when the workflow is
    cancelled or fails.
```


message_passing/message_handler_waiting_compensation_cleanup/workflows.py
```diff
@@ -0,0 +1,121 @@
...
+            # handlers to finish in a `finally` block. The reason is that other
+            # exception types will cause a Workflow Task failure, in which case
+            # we do *not* want to wait for message handlers to finish.
+            result = await self._run(input)
+            await workflow.wait_condition(workflow.all_handlers_finished)
+            return result
+        except (
+            asyncio.CancelledError,
+            workflow.ContinueAsNewError,
+            exceptions.FailureError,
```
drewhoskins-temporal, line 45: I wonder if we add a wrapper here so that one can do like:
```
except e:
   if is_workflow_exit_exception(e):
     await workflow.wait_condition(workflow.all_handlers_finished)
   raise e
```

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
```diff
@@ -0,0 +1,121 @@
...
+        # prematurely in the same tick of the event loop. If the Update has
+        # completed, return the Update result to the caller, whether or not the
+        # Workflow is exiting.
+        if (
+            update_task in first_completed
+            or input.on_premature_workflow_exit == OnWorkflowExitAction.CONTINUE
+        ):
+            return await update_task
+        else:
+            await self._my_update_compensation_and_cleanup()
```
cretz, line 77: Usually when people refer to compensations, they want to clean up on any failure, not just on workflow complete. This compensation is not run on update task failure.

I would recommend calling `asyncio.Future.set_exception` for the workflow exit exception, and changing to `return_when=asyncio.FIRST_EXCEPTION`, and put this compensation in an `except` before re-raising. This also may clean up your future generic and let you assume the successful result of wait is a single-item array of the update task

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
```diff
@@ -0,0 +1,121 @@
...
+        self.workflow_exit_exception: asyncio.Future[BaseException] = asyncio.Future()
+
+    @workflow.run
+    async def run(self, input: WorkflowInput) -> str:
+        try:
+            # 👉 Use this `try...except` style, instead of waiting for message
+            # handlers to finish in a `finally` block. The reason is that other
+            # exception types will cause a Workflow Task failure, in which case
+            # we do *not* want to wait for message handlers to finish.
+            result = await self._run(input)
```
cretz, line 39: I think splitting the run logic to half at the top of the class and have way at the bottom makes the sample hard to read, consider inlining

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
```diff
@@ -0,0 +1,121 @@
...
+
+        (1) In the case where the workflow finishes successfully, the worker
+            would get an UnfinishedUpdateHandlersWarning (TMPRL1102) if the main
+            workflow task didn't wait for it to finish.
+
+        (2) In the other cases (failure, cancellation, and continue-as-new), the
+            premature workflow exit will occur before the update is finished.
+        """
+        # Ignore: implementation detail specific to this sample
+        self._update_started = True
```
cretz, line 94: We haven't traditionally encouraged users to prefix `_` to private workflow things since the only users that get an instance of the workflow are the implementers. Up to you, but it's not a common pattern you'll see elsewhere in our docs/samples (as opposed to cases where you do want to make it clear to people holding an instance that certain methods are off limits).

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
```diff
@@ -0,0 +1,121 @@
...
+    @workflow.update
+    async def my_update(self, input: UpdateInput) -> str:
+        """
+        This update handler demonstrates how to handle the situation where the
+        main Workflow method exits prematurely. In that case we perform
+        compensation/cleanup, and fail the Update. The Update caller will get a
+        WorkflowUpdateFailedError.
+        """
+        # Coroutines must be wrapped in tasks in order to use workflow.wait.
+        update_task = asyncio.Task(self._my_update())
```
cretz, line 60:
```suggestion
        update_task = asyncio.create_task(self._my_update())
```

This is the recommended way to create a task (starts it in the background). Per https://docs.python.org/3/library/asyncio-task.html#asyncio.Task:

> Use the high-level [asyncio.create_task()](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task) function to create Tasks, or the low-level [loop.create_task()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_task) or [ensure_future()](https://docs.python.org/3/library/asyncio-future.html#asyncio.ensure_future) functions. Manual instantiation of Tasks is discouraged.

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
```diff
@@ -0,0 +1,121 @@
...
+        workflow.logger.info(
+            "performing update handler compensation and cleanup operations"
+        )
+
+    async def _run(self, input: WorkflowInput) -> str:
+        # Ignore this method unless you are interested in the implementation
+        # details of this sample.
+
+        # Wait until handlers started, so that we are demonstrating that we wait for them to finish.
+        await workflow.wait_condition(lambda: getattr(self, "_update_started", False))
```
cretz, line 111:
```suggestion
        await workflow.wait_condition(lambda: self._update_started)
```

And would recommend initializing it to `False` in the constructor

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
cretz, line 77: May want to demonstrate running an activity in the cleanup since that's the most common approach. May also want to cancel the update's activity first (whether you want to wait for update cancellation is up to you).

Also, one of the problems that plagues compensation writers is what to do with an error on compensation. Will leave it up to you, but at the least if we want to leave it like this, should acknowledge in a comment above this that we are intentionally letting raised compensation exceptions swallow the existing one.

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
```diff
@@ -0,0 +1,121 @@
...
+        if (
+            update_task in first_completed
+            or input.on_premature_workflow_exit == OnWorkflowExitAction.CONTINUE
+        ):
+            return await update_task
+        else:
+            await self._my_update_compensation_and_cleanup()
+            raise exceptions.ApplicationError(
+                f"The update failed because the workflow run exited: {await self.workflow_exit_exception}"
+            )
```
cretz, line 80:
```suggestion
            raise exceptions.ApplicationError(
                "The update failed because the workflow run exited"
            ) from await self.workflow_exit_exception
```

This retains the stack trace and cause chain and such (though if/when changing the logic to use future exception this logic may be a bit clearer)

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
```diff
@@ -0,0 +1,121 @@
...
+        # details of this sample.
+
+        # Wait until handlers started, so that we are demonstrating that we wait for them to finish.
+        await workflow.wait_condition(lambda: getattr(self, "_update_started", False))
+        if input.exit_type == WorkflowExitType.SUCCESS:
+            return "workflow-result"
+        elif input.exit_type == WorkflowExitType.CONTINUE_AS_NEW:
+            workflow.continue_as_new(WorkflowInput(exit_type=WorkflowExitType.SUCCESS))
+        elif input.exit_type == WorkflowExitType.FAILURE:
+            raise exceptions.ApplicationError("deliberately failing workflow")
```
cretz, line 117: Inconsistent with whether you capitalize exception messages (in other places we do)

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
```diff
@@ -0,0 +1,121 @@
...
+        self._update_started = True
+
+        await workflow.execute_activity(
+            my_activity, start_to_close_timeout=timedelta(seconds=10)
+        )
+        return "update-result"
+
+    async def _my_update_compensation_and_cleanup(self):
+        workflow.logger.info(
+            "performing update handler compensation and cleanup operations"
```
cretz, line 103: Pedantic, but in other places we capitalize logs

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
dandavison: `asyncio.CancelledError` can occur in many different contexts and should only be a "workflow exit" exception in a small subset of those contexts. I'm not sure about the need for a utility method, and I don't think we would want a function `is_workflow_exit_exception` to simply return `True` for an instance of `asyncio.CancelledError`.

cretz: > asyncio.CancelledError can occur in many different contexts and should only be considered a "workflow exit" exception in a small subset of those contexts.

It is currently always considered a workflow exit exception when raised from the workflow (see [here](https://github.com/temporalio/sdk-python/blob/2e6d742ce1531741c0f953a311689bd72781b720/temporalio/worker/_workflow_instance.py#L1834-L1836)). It is never treated as a task failure. IMO another way to phrase this helper instead of "is workflow exit exception" is more like "is not workflow task failure exception", but the former reads better as a function name.

cretz: > I think we may want to make a change to the SDK to make the failure_exception_types set at worker and/or workflow level available.

I think the helper would be enough for now. I don't think a user using those collections directly value (we may change what we consider a workflow exit exception, but yes currently it's Temporal failure, asyncio cancelled error, or an instance of a type in one of those two collections).

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
cretz: Arguably, it's just a bit hard to follow to have to jump around the file for this very simple primary workflow logic. But not a big deal.

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
cretz: I think it should be in view and IMO we should encourage explicit attributes over string-based ones, even for control-based attributes like this one. But it's not a big deal.

message_passing/message_handler_waiting_compensation_cleanup/workflows.py
dandavison: This was a very helpful comment (and certainly a bug that it wasn't handling failures in the update handler). I've rewritten the "control plane" logic here (partly in collaboration with @cretz offline) and am writing tests for the sample.

dandavison: Agreed, done.

message_passing/waiting_for_handlers_and_compensation/workflows.py
```diff
@@ -0,0 +1,180 @@
...
+        # that coroutines must be wrapped in tasks in order to use
+        # workflow.wait.
+        update_task = asyncio.create_task(self._my_update())
+
+        # 👉 "Race" the workflow_exit future against the handler's own application
+        # logic. Always use `workflow.wait` instead of `asyncio.wait` in
+        # Workflow code: asyncio's version is non-deterministic.
+        await workflow.wait(  # type: ignore
+            [update_task, self.workflow_exit], return_when=asyncio.FIRST_EXCEPTION
+        )
```
drewhoskins-temporal, line 102: This makes sense, and looks like a pain, but I'm not sure how often people want to cancel in-process update handlers in response to a workflow exit as opposed to simply waiting for the handlers to complete.
My guess is that most people are happy to wait because such cases are not particularly latency sensitive.
It may be okay that this is more of a power user struggle.

message_passing/waiting_for_handlers_and_compensation/workflows.py
```diff
@@ -0,0 +1,180 @@
...
+
+        It ensures that:
+        - Compensation/cleanup is always performed when appropriate
+        - The update caller gets the update result, or WorkflowUpdateFailedError
+        """
+        # 👉 As with the main workflow method, the update application logic is
+        # implemented in a separate method in order to separate "platform-level"
+        # error-handling and compensation concerns from application logic. Note
+        # that coroutines must be wrapped in tasks in order to use
+        # workflow.wait.
```
drewhoskins-temporal, line 94: nit: This struck me as abstract.  Lines 97-102 are what helped me understand what's going on, so I would probably drop this.

message_passing/waiting_for_handlers_and_compensation/workflows.py
```diff
@@ -0,0 +1,180 @@
...
+            else:
+                # 👉 The main workflow method exited prematurely due to an
+                # error, and this happened before the update finished. Fail the
+                # update with the workflow exception as cause.
+                raise exceptions.ApplicationError(
+                    "The update failed because the workflow run exited"
+                ) from cast(BaseException, self.workflow_exit.exception())
+        # 👉 Catch BaseException since asyncio.CancelledError does not inherit
+        # from Exception.
+        except BaseException as e:
```
drewhoskins-temporal, line 121: I haven't thought deeply, but I'm wondering if using a child workflow would work better for use cases in which there are compensations.  Would it be more straightforward?  Would it work better, e.g. by letting the workflow exit and letting the child workflow finish up the compensations?
Obviously that's extra latency which some might not be able to afford.  And it's a lot of code, though it might be easier than this.

message_passing/waiting_for_handlers_and_compensation/workflows.py
```diff
@@ -0,0 +1,180 @@
...
+            if is_workflow_exit_exception(e):
+                try:
+                    await self.my_update_compensation()
+                except BaseException as e:
+                    raise exceptions.ApplicationError(
+                        "Update compensation failed"
+                    ) from e
+            raise
+
+    async def my_update_compensation(self):
```
drewhoskins-temporal, line 131: It'd be cool if folks could register cancellation handlers with an annotation similar to validators.

message_passing/waiting_for_handlers_and_compensation/workflows.py
```diff
@@ -0,0 +1,180 @@
...
+            return "workflow-result"
+        elif input.exit_type == WorkflowExitType.FAILURE:
+            raise exceptions.ApplicationError("deliberately failing workflow")
+        elif input.exit_type == WorkflowExitType.CANCELLATION:
+            # Block forever; the starter will send a workflow cancellation request.
+            await asyncio.Future()
+        raise AssertionError("unreachable")
+
+
+def is_workflow_exit_exception(e: BaseException) -> bool:
```
drewhoskins-temporal, line 177: 👍 , This feels like a concept worth highlighting (and maybe adding a helper?) that I think lots of users would benefit from.  Good to take every opportunity to teach users about which exceptions fail the workflow and the fact that some don't fail it (and that this is a good thing), and then show how that intersects with cleanup code like this.

drewhoskins-temporal (commented):
Feels to me like there's two main things going on here.
* cleaning up the workflow's handlers
* aborting updates and compensating for them.

Both seem valid.  The first feels mainstream, the second power user (though we may start to get feedback telling us otherwise).

I wonder if they should be separate samples, or at least more clearly frame this one so people can just pick out what they need.

I can imagine pretty achievable framework improvements we could do for both scenarios.  Given that we've heard a lot more feedback about the first scenario, that'd be my guess as to where to invest for now.

message_passing/waiting_for_handlers_and_compensation/workflows.py
dandavison: Agreed, done.

message_passing/waiting_for_handlers_and_compensation/workflows.py
dandavison: I'm not sure I agree with this. The point isn't the latency -- it's that if the workflow is exiting due to a failure/cancellation then the update may need to perform a rollback / cancellation. So they can't "simply wait for the handler to complete" because the handler has to learn whether it should proceed to the end of its happy path, or perform rollback / compensation.

message_passing/waiting_for_handlers_and_compensation/workflows.py
dandavison: Right, I think we all agreed on that. https://github.com/temporalio/features/issues/589

message_passing/waiting_for_handlers_and_compensation/workflows.py
dandavison: Interesting idea! Make a note of it?

dandavison:
> I wonder if they should be separate samples

Yes, I made them into two separate samples: 

```markdown
# Waiting for message handlers

This workflow demonstrates how to wait for signal and update handlers to
finish in the following circumstances:

- Before a successful return
- On failure
- On cancellation
```

```markdown
# Waiting for message handlers, and performing compensation and cleanup in message handlers

This sample demonstrates how to do the following:

1. Ensure that all update/signal handlers are finished before a successful
    workflow return, and on workflow cancellation and failure.
2. Perform compensation/cleanup in an update handler when the workflow is
    cancelled or fails.
```

This should be ready to merge now.

cretz (approved):
Nothing blocking

message_passing/waiting_for_handlers/workflows.py
```diff
@@ -0,0 +1,95 @@
...
+    WorkflowExitType,
+    WorkflowInput,
+    WorkflowResult,
+)
+from message_passing.waiting_for_handlers.activities import (
+    activity_executed_by_update_handler,
+)
+
+
+def is_workflow_exit_exception(e: BaseException) -> bool:
```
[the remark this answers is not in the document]
dandavison: Yep, it says that in the comment:

```python
    # 👉 If you have set additional failure_exception_types you should also
    # check for these here.
```

and we have https://github.com/temporalio/features/issues/589 for the utility.

message_passing/waiting_for_handlers/starter.py
```diff
@@ -0,0 +1,69 @@
...
+            f"    🔴 caught exception while waiting for workflow result: {e}: {e.__cause__ or ''}"
+        )
+
+
+async def main():
+    for exit_type in [
+        WorkflowExitType.SUCCESS,
+        WorkflowExitType.FAILURE,
+        WorkflowExitType.CANCELLATION,
+    ]:
```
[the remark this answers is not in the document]
dandavison: The purpose of this sample is partly to teach users about what errors they'll see under different sorts of workflow exit. See the README:

```
workflow exit type: SUCCESS
    🟢 caller received update result
    🟢 caller received workflow result


workflow exit type: FAILURE
    🟢 caller received update result
    🔴 caught exception while waiting for workflow result: Workflow execution failed: deliberately failing workflow


workflow exit type: CANCELLATION
    🟢 caller received update result
```

message_passing/waiting_for_handlers/workflows.py
```diff
@@ -0,0 +1,95 @@
...
+            # handlers to finish.
+            result = await self._my_workflow_application_logic(input)
+            await workflow.wait_condition(workflow.all_handlers_finished)
+            return result
+        # 👉 Catch BaseException since asyncio.CancelledError does not inherit
+        # from Exception.
+        except BaseException as e:
+            if is_workflow_exit_exception(e):
+                await workflow.wait_condition(workflow.all_handlers_finished)
+            raise
```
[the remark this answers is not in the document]
dandavison: I think that showing them how to wait for the handler when the workflow fails is useful since it points them towards how to implement graceful cleanup of a handler execution.

PR #145, opened 2024-10-05
Update hello_search_attributes.py with new typed search attribute syntax

accupham:
Old example used deprecated syntax

## What was changed
Update hello_search_attributes.py with new typed search attribute syntax

## Why?
Old example used deprecated syntax

### DIFF
--- a/hello/hello_search_attributes.py
+++ b/hello/hello_search_attributes.py
@@ -3,6 +3,7 @@ import asyncio
 from temporalio import workflow
 from temporalio.client import Client
 from temporalio.worker import Worker
+from temporalio.common import SearchAttributeKey
 
 
 @workflow.defn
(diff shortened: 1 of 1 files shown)


hello/hello_search_attributes.py
```diff
@@ -28,23 +31,26 @@ async def main():
         # While the worker is running, use the client to start the workflow.
         # Note, in many production setups, the client would be in a completely
         # separate process from the worker.
+        custom_keyword_field = SearchAttributeKey.for_keyword("CustomKeywordField")
```
cretz, line 34: Could even make this a global if you'd like and reuse it in the workflow too. At the very least, define it at the top of this function instead of redefining within the function.

cretz:
Thanks! Left a comment. May also need to sign the CLA.

cretz:
May need to run `poe format`.

PR #152, opened 2024-11-21
Fix polling sample 2

dandavison:
Minor modification of https://github.com/temporalio/samples-python/pull/150. Use a global variable to separate implementation details from the user-facing lesson.


cretz (approved):
LGTM but may want to wait on @drewhoskins. As for CI failure, just changing `macos-12` to `macos-13` should fix it, feel free to do inside this PR, doesn't necessarily have to be a separate one.

drewhoskins-temporal (approved):

PR #153, opened 2024-12-08
Cluster manager: use update to wait for cluster to start

dandavison:
This PR changes our cluster manager sample as follows:

- Previously, the workflow was started, but the "cluster itself did not start" until a signal is sent in.
- Now, the workflow automatically does the "start cluster" operations on workflow start. An update is available for the user to wait until the cluster is ready to be used.

The reason for this change is that it highlights two styles of update usage:

1. The update accepts some input and does something with it (assign nodes to cluster, or execute an activity in general)
2. The update is used to wait until a certain stage (cluster started, early return available, etc)

In case (2) the workflow should generally be written such that it will orchestrate its work whether or not the update is sent, and that's what this PR does.

Also:
- switch to use `@workflow.init`
- Use `hash(node_id) % 5` instead of requiring node IDs to parse as ints

### DIFF
--- a/message_passing/safe_message_handlers/workflow.py
+++ b/message_passing/safe_message_handlers/workflow.py
@@ -74,13 +74,14 @@ class ClusterManagerWorkflow:
             self.state = ClusterManagerState()
 
         if input.test_continue_as_new:
-            self.max_history_length = 120
+            self.max_history_length: Optional[int] = 120
             self.sleep_interval_seconds = 1
+        else:
+            self.max_history_length = None
+            self.sleep_interval_seconds = 600
 
         # Protects workflow state from interleaved access
         self.nodes_lock = asyncio.Lock()
-        self.max_history_length: Optional[int] = None
-        self.sleep_interval_seconds: int = 600
 
     @workflow.update
     async def wait_until_cluster_started(self) -> ClusterManagerState:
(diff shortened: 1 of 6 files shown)


message_passing/safe_message_handlers/workflow.py
```diff
@@ -65,18 +66,26 @@ class ClusterManagerAssignNodesToJobResult:
...
         self.max_history_length: Optional[int] = None
         self.sleep_interval_seconds: int = 600
 
-    @workflow.signal
-    async def start_cluster(self) -> None:
-        self.state.cluster_started = True
-        self.state.nodes = {str(k): None for k in range(25)}
-        workflow.logger.info("Cluster started")
+    @workflow.update
+    async def wait_until_cluster_started(self) -> ClusterManagerState:
```
dandavison, line 86: Is this the right name for the update?

cretz (approved):

message_passing/safe_message_handlers/workflow.py
dandavison: Can't think of a better name. It matches the name of the state variable:
```python
    async def wait_until_cluster_started(self) -> ClusterManagerState:
        await workflow.wait_condition(lambda: self.state.cluster_started)
```

PR #154, opened 2024-12-10
STYLE: fix violation of the DRY principle

drveles:
## What was changed
The code has been updated in accordance with the [DRY principle](https://ru.wikipedia.org/wiki/Don’t_repeat_yourself).

## Why?
It is much smaller, more concise and faster to read.

1. Closes 

2. How was this tested:
Standard tests must pass.

3. Any docs updates needed?
No. The logic has not been changed.

### DIFF
--- a/hello/hello_activity_choice.py
+++ b/hello/hello_activity_choice.py
@@ -59,39 +59,22 @@ class PurchaseFruitsWorkflow:
         ordered: List[str] = []
         for item in list.items:
             if item.fruit is Fruit.APPLE:
-                ordered.append(
-                    await workflow.execute_activity(
-                        order_apples,
-                        item.amount,
-                        start_to_close_timeout=timedelta(seconds=5),
-                    )
-                )
+                order_function = order_apples
             elif item.fruit is Fruit.BANANA:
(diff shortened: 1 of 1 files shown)


cretz (approved):
Sometimes in samples we intentionally copy code as a demonstration of what a user might do if they had different code for each conditional. For example, if we really wanted to properly reuse code, we wouldn't have 4 separately named activities, we'd have one `order_fruit` activity that accepted a fruit type.

But I am ok with this. @dandavison - opinion?

dandavison:
Nice cleanup @drveles. I agree -- while the activity calls are all the same it's appropriate for them to be written without repetition.

hello/hello_activity_choice.py
```diff
@@ -59,39 +59,22 @@ async def run(self, list: ShoppingList) -> str:
         ordered: List[str] = []
         for item in list.items:
```
dandavison, line 60: A tiny thing, not related to this PR, but we should avoid using `list` as a variable name in Python.

drveles:
Thank you. 
I will probably correct the code according to the remark about the `list'.

PR #156, opened 2024-12-12
Update-with-start: shopping cart

dandavison:
- The Workflow represents a Shopping Cart in an e-commerce application, and
update-with-start is used to add items to the cart, receiving back the updated cart subtotal.

- Cannot be merged until UwS is released


README.md
```diff
@@ -62,13 +62,14 @@ Some examples require extra dependencies. See each sample's directory for specif
 * [encryption](encryption) - Apply end-to-end encryption for all input/output.
 * [gevent_async](gevent_async) - Combine gevent and Temporal.
 * [langchain](langchain) - Orchestrate workflows for LangChain.
-* [message-passing introduction](message_passing/introduction/) - Introduction to queries, signals, and updates.
+* [message_passing/introduction](message_passing/introduction/) - Introduction to queries, signals, and updates.
+* [message_passing/safe_message_handlers](message_passing/safe_message_handlers/) - Safely handling updates and signals.
+* [message_passing/update_with_start/lazy_init](message_passing/update_with_start/lazy_init/) - Use update-with-start to update a Shopping Cart, starting if it it does not exist.
```
cretz, line 67:
```suggestion
* [message_passing/update_with_start/lazy_initialization](message_passing/update_with_start/lazy_initialization/) - Use update-with-start to update a Shopping Cart, starting if it it does not exist.
```

message_passing/update_with_start/lazy_initialization/README.md
```diff
@@ -0,0 +1,16 @@
+# Update With Start: Lazy init
+
+This sample illustrates the use of update-with-start to send Updates to a Workflow, starting the Workflow if
+it is not running yet. The Workflow represents a Shopping Cart in an e-commerce application, and
+update-with-start is used to add items to the cart, receiving back the updated cart subtotal.
+
+Run the following from this directory:
```
cretz, line 7:
```suggestion
To run, first see [README.md](../../README.md) for prerequisites. Then, run the following from this directory:
```

We have tried in the past to have samples refer to the README for prereqs (namely starting a local server), though maybe some samples are not doing that lately. Unsure if we also want to document how to enable multi-op since that will become stale and we may forget to come back and update the instructions.

message_passing/update_with_start/lazy_initialization/workflows.py
```diff
@@ -0,0 +1,51 @@
...
+            raise ApplicationError(
+                f"Item unavailable: {item}",
+            )
+        self.items.append((item, Decimal(price)))
+
+        return str(sum(price for _, price in self.items) or Decimal("0.00"))
+
+    @workflow.signal
+    def checkout(self):
+        self.order_submitted = True
```
cretz, line 51: Hrmm, I wonder if we should add a validator that disallows `add_item` if this is true. While this workflow is safe and doesn't need it, it could be a good demo, and would help users if they ever had a step between the wait condition and the return.

message_passing/update_with_start/lazy_initialization/workflows.py
```diff
@@ -0,0 +1,51 @@
...
+from decimal import Decimal
+from typing import List, Tuple
+
+from temporalio import workflow
+from temporalio.exceptions import ApplicationError
+
+from message_passing.update_with_start.lazy_initialization.activities import (
+    ShoppingCartItem,
+    get_price,
+)
```
cretz, line 11:
```suggestion
with workflow.unsafe.imports_passed_through():
    from message_passing.update_with_start.lazy_initialization.activities import (
        ShoppingCartItem,
        get_price,
    )
```

Need to encourage passing through all activity references and externally defined models instead of reloading them every workflow (for the samples this technically works because the activity doesn't do anything externally meaningful)

message_passing/update_with_start/lazy_initialization/README.md
```diff
@@ -0,0 +1,16 @@
+# Update With Start: Lazy init
```
yuandrew, line 1: Can you add a little blurb as to what "lazy init" means here? Does lazy init here refer to how we're using update_with_start to initialize and start the workflow?

message_passing/update_with_start/lazy_initialization/workflows.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+        return FinalizedOrder(
+            id=workflow.info().workflow_id,
+            items=[(item, str(price)) for item, price in self.items],
+            total=str(sum(price for _, price in self.items) or Decimal("0.00")),
+        )
+
+    @workflow.update
+    async def add_item(self, item: ShoppingCartItem) -> str:
+        price = await get_price(item)
```
cretz, line 40: I think you want this to be an activity invocation

message_passing/update_with_start/lazy_initialization/workflows.py
dandavison: I don't think I've ever once remembered to do this.

message_passing/update_with_start/lazy_initialization/workflows.py
dandavison: Thanks! So (a) too much Typescript, and (b) this is a good example of a mistake that an AI linter could catch but traditional analysis wouldn't.

message_passing/update_with_start/lazy_initialization/workflows.py
dandavison: Good suggestion. Done.

message_passing/update_with_start/lazy_initialization/README.md
dandavison: Done!

message_passing/update_with_start/lazy_initialization/workflows.py
dandavison: Fixed

dandavison:
@cretz @yuandrew thanks for your reviews -- this should be ready now.

message_passing/update_with_start/lazy_initialization/starter.py
```diff
@@ -0,0 +1,68 @@
...
+        task_queue="uws",
+    )
+    try:
+        price = await temporal_client.execute_update_with_start_workflow(
+            ShoppingCartWorkflow.add_item,
+            ShoppingCartItem(sku=item_id, quantity=quantity),
+            start_workflow_operation=start_op,
+        )
+    except WorkflowUpdateFailedError:
+        price = None
```
cretz, line 45: May want a bit more of a conditional to confirm it's the failure you expect instead of swallowing every update failure ever. In .NET I checked the application failure's error type: https://github.com/temporalio/samples-dotnet/blob/86a49887aaa668beb3cadb4fa862574f3295fcc6/src/UpdateWithStartLazyInit/Program.cs#L85-L90

message_passing/update_with_start/lazy_initialization/__init__.py
```diff
@@ -0,0 +1 @@
+TASK_QUEUE = "uws"
```
cretz, line 1: I think the task queue should be named specific to the sample so it doesn't clash with others (also, this constant isn't used in `starter.py` as might be expected)

cretz (approved):
Nothing blocking

message_passing/update_with_start/lazy_initialization/__init__.py
dandavison: done

PR #160, opened 2025-01-27
sleep-for-days sample

THardy98:
This sample demonstrates how to use Temporal to run a workflow that periodically sleeps for a number of days.

### What was changed

added sleep-for-days sample

### Any docs updates needed?
I believe so @webchick


sleep_for_days/workflows.py
```diff
@@ -0,0 +1,26 @@
...
+class SleepForDaysWorkflow:
+
+    def __init__(self) -> None:
+        self.is_complete = False
+
+    @workflow.run
+    async def run(self, input: SleepForDaysInput) -> str:
+        while(not self.is_complete):
+            await workflow.execute_activity(send_email, SendEmailInput(f"{input.numOfDays} until the next email"), start_to_close_timeout=timedelta(seconds=10))
+            await asyncio.sleep(timedelta(days=input.numOfDays).seconds)
```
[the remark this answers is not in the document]
THardy98: Done.

sleep_for_days/workflows.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def __init__(self) -> None:
+        self.is_complete = False
+
+    @workflow.run
+    async def run(self, input: SleepForDaysInput) -> str:
+        while(not self.is_complete):
+            await workflow.execute_activity(send_email, SendEmailInput(f"{input.numOfDays} until the next email"), start_to_close_timeout=timedelta(seconds=10))
+            await workflow.wait([
+                asyncio.create_task(workflow.sleep(input.numOfDays * 24 * 60 * 60)),
```
Sushisource, line 22: This would look nicer as a `timedelta` which `sleep` accepts rather than doing all this multiplication

sleep_for_days/workflows.py
```diff
@@ -0,0 +1,29 @@
...
+
+@workflow.defn(name="SleepForDaysWorkflow")
+class SleepForDaysWorkflow:
+
+    def __init__(self) -> None:
+        self.is_complete = False
+
+    @workflow.run
+    async def run(self, input: SleepForDaysInput) -> str:
+        while(not self.is_complete):
```
dandavison, line 19: This looks a bit JS-y :)

sleep_for_days/workflows.py
```diff
@@ -0,0 +1,29 @@
+import asyncio
+from datetime import timedelta
+from temporalio import workflow
+from dataclasses import dataclass
+from sleep_for_days.activities import send_email, SendEmailInput
+
+@dataclass
+class SleepForDaysInput:
+    numOfDays: int
```
dandavison, line 9: Let's use snake_case for Python

sleep_for_days/workflows.py
```diff
@@ -0,0 +1,29 @@
...
+
+@workflow.defn(name="SleepForDaysWorkflow")
```
dandavison, line 11: No need for name if it's the same as the class name

sleep_for_days/workflows.py
dandavison: What is the while loop needed for?

sleep_for_days/workflows.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self) -> None:
+        self.is_complete = False
+
+    @workflow.run
+    async def run(self, input: SleepForDaysInput) -> str:
+        while(not self.is_complete):
+            await workflow.execute_activity(send_email, SendEmailInput(f"{input.numOfDays} until the next email"), start_to_close_timeout=timedelta(seconds=10))
+            await workflow.wait([
+                asyncio.create_task(workflow.sleep(input.numOfDays * 24 * 60 * 60)),
+                asyncio.create_task(workflow.wait_condition(lambda: self.is_complete))
```
dandavison, line 23: It would be simpler to use the timeout argument of `wait_condition` instead of racing these two tasks.

sleep_for_days/workflows.py
```diff
@@ -0,0 +1,29 @@
+import asyncio
+from datetime import timedelta
+from temporalio import workflow
+from dataclasses import dataclass
+from sleep_for_days.activities import send_email, SendEmailInput
```
dandavison, line 5: In workflow code we import activity code with the sandbox disabled, as follows:

```
    from ...
```

It's hard to remember to do it though... would be good to improve that.

dandavison:
One more thing -- can you add a test? (I'd like us to have tests for all samples)

sleep_for_days/workflows.py
THardy98: Ah I noticed that with the other samples, but the `unsafe` scared me so I imported it normally.

THardy98:
> One more thing -- can you add a test? (I'd like us to have tests for all samples)

Added a test

sleep_for_days/workflows.py
```diff
@@ -0,0 +1,37 @@
...
+class SleepForDaysWorkflow:
+    def __init__(self) -> None:
+        self.is_complete = False
+
+    @workflow.run
+    async def run(self) -> str:
+        while not self.is_complete:
+            await workflow.execute_activity(
+                send_email,
+                SendEmailInput("30 until the next email"),
```
dandavison, line 21:
```suggestion
                SendEmailInput("30 days until the next email"),
```

tests/sleep_for_days/workflow_test.py
```diff
@@ -0,0 +1,53 @@
...
+            # Time-skip 3 days.
+            await env.sleep(timedelta(days=90))
+            # Expect 3 more activity calls.
+            assert num_activity_executions == 4
+            # Send the signal to complete the workflow.
+            await handle.signal(SleepForDaysWorkflow.complete)
+            # Expect no more activity calls to have been made - workflow is complete.
+            assert num_activity_executions == 4
+            # Expect more than 90 days to have passed.
+            assert (await env.get_current_time() - start_time) > timedelta(days=90)
```
dandavison, line 53: Nice!

tests/sleep_for_days/workflow_test.py
```diff
@@ -0,0 +1,53 @@
...
+    async def send_email_mock(input: SendEmailInput) -> str:
+        nonlocal num_activity_executions
+        num_activity_executions += 1
+        return input.email_msg
+
+    async with await WorkflowEnvironment.start_time_skipping() as env:
+        # if env.supports_time_skipping:
+        #     pytest.skip(
+        #         "Java test server: https://github.com/temporalio/sdk-java/issues/1903"
+        #     )
```
dandavison, line 26:
```suggestion
```

sleep_for_days/workflows.py
```diff
@@ -0,0 +1,37 @@
...
+            )
+            await workflow.wait(
+                [
+                    asyncio.create_task(workflow.sleep(timedelta(days=30))),
+                    asyncio.create_task(
+                        workflow.wait_condition(lambda: self.is_complete)
+                    ),
+                ],
+                return_when=asyncio.FIRST_COMPLETED,
+            )
```
dandavison, line 32: It's a little unfortunate that `asyncio` makes this so cumbersome, I'd like us not to have to explicitly schedule the coroutines, but that's how the stdlib is.

sleep_for_days/README.md
```diff
@@ -0,0 +1,18 @@
+# Sleep for Days
+
+This sample demonstrates how to use Temporal to run a workflow that periodically sleeps for a number of days.
```
dandavison, line 3:
```suggestion
This sample demonstrates how to create a Temporal workflow that runs forever, sending an email every 30 days.
```

sleep_for_days/activities.py
```diff
@@ -0,0 +1,18 @@
...
+    email_msg: str
+
+
+@activity.defn()
+async def send_email(input: SendEmailInput) -> str:
+    """
+    A stub Activity for sending an email.
+    """
+    result = f"Email message: {input.email_msg}, sent"
+    print(result)
```
dandavison, line 17: Arguably since this output is coming from the worker process, it should use logging. (I think the print statement is fine in the client/starter process.)

sleep_for_days/workflows.py
```diff
@@ -0,0 +1,37 @@
...
```
cretz, line 32: `wait_condition` has a timeout which I think would have been clearer here

PR #162, opened 2025-02-05
Change minimum Python to 3.9 and add Trio sample

### LINKED ISSUE #161: Sample request: Trio
### Describe the solution you'd like

Demonstrate an `async def` activity running in trio-backed asyncio using trio utilities. There is a user report that trio-asyncio interop is not working. May just need to demonstrate https://trio-asyncio.readthedocs.io/en/latest/usage.html#calling-trio-from-asyncio.

cretz:
## What was changed

* Changed minimum Python version to 3.9 
  * We'd have to do this anyways when we upgrade to next Python release, just didn't do the 3.13 update in CI
  * I'll update the required CI checks to be 3.9 just before merge of this PR
* Added sample for Trio demonstrating using Temporal in a codebases that have fully adopted Trio

## Checklist

1. Closes #161

### DIFF
--- a/trio_async/activities.py
+++ b/trio_async/activities.py
@@ -17,9 +17,9 @@ async def say_hello_activity_async(name: str) -> str:
     await asyncio.sleep(0.1)
 
     # Now Trio. We have to invoke the function separately decorated.
-    # Unfortunately the @trio_as_aio decorator doesn't do a proper functools
-    # wrap so it doesn't respond to things like __name__ that @activity.defn
-    # needs.
+    # We cannot use the @trio_as_aio decorator on the activity itself because
+    # it doesn't use functools wrap or similar so it doesn't respond to things
+    # like __name__ that @activity.defn needs.
     return await say_hello_in_trio_from_asyncio(name)
 
 


trio_async/activities.py
```diff
@@ -0,0 +1,51 @@
...
+    # and Trio primitives can be used
+
+    # First asyncio
+    activity.logger.info("Sleeping in asyncio")
+    await asyncio.sleep(0.1)
+
+    # Now Trio. We have to invoke the function separately decorated.
+    # Unfortunately the @trio_as_aio decorator doesn't do a proper functools
+    # wrap so it doesn't respond to things like __name__ that @activity.defn
+    # needs.
```
dandavison, line 22: Let's not digress into criticizing 3rd party libraries -- they may well have a reason for their implementation/design decisions.

Let's either delete this, or clarify what it is that users cannot do that they might expect to be able to (it's not quite clear to me).

```suggestion
```

trio_async/starter.py
```diff
@@ -0,0 +1,28 @@
+import logging
+
+import trio_asyncio
+from temporalio.client import Client
+
+from trio_async import workflows
+
+
+@trio_asyncio.aio_as_trio  # Note this decorator which allows asyncio primitives
```
dandavison, line 9: Could you expand this comment slightly, are we basically saying

> without this decorator you wouldn't be able to use anything from the `asyncio` module.

?

trio_async/workflows.py
```diff
@@ -0,0 +1,30 @@
...
+    from trio_async.activities import say_hello_activity_async, say_hello_activity_sync
+
+
+@workflow.defn
+class SayHelloWorkflow:
+    @workflow.run
+    async def run(self, name: str) -> list[str]:
+        # Workflows don't use default asyncio event loop or Trio, they use a
+        # custom event loop. Therefore Trio primitives should never be used in a
+        # workflow, only asyncio helpers (which delegate to the custom loop).
```
dandavison, line 15: These comments are very helpful. Two teaks: if "primitives" and "helpers" are the same thing can we use the same word, and can you give an example so that it's really clear what we can and can't do. Something like this:

```suggestion
        # custom event loop. Therefore Trio primitives such as XXX should never be used in a
        # workflow, only functions and classes from the asyncio module, (which delegate to the custom loop).
```

trio_async/worker.py
```diff
@@ -0,0 +1,66 @@
...
+import sys
+
+import trio_asyncio
+from temporalio.client import Client
+from temporalio.worker import Worker
+
+from trio_async import activities, workflows
+
+
+@trio_asyncio.aio_as_trio  # Note this decorator which allows asyncio primitives
```
dandavison, line 13: As above, would be good to expand and clarify this. Are we saying

> without this decorator you wouldn't be able to use anything from the `asyncio` module.

?

dandavison (approved):
Nice, a few comments aimed at making the comments maximally understandable.

trio_async/activities.py
cretz: It's explaining why we can't use the decorator on activities. I can remove the "Unfortunately" bit but it's not criticizing the library, it's an unfortunate situation for those relying on `__name__`.

trio_async/starter.py
cretz: Yes, "allows asyncio primitives" was meant to imply "needed to allow asyncio primitives", otherwise if it wasn't needed we could remove the decorator. I can rephrase if needed (though don't want to effectively re-document https://trio-asyncio.readthedocs.io/en/latest/usage.html#calling-asyncio-from-trio).

trio_async/worker.py
cretz: Yes, I can adjust. I do not believe it is true that it's _anything_ from the asyncio module because I didn't go over everything in that module.

trio_async/activities.py
dandavison: You'd need to remove "proper" also. But our samples repos just need to be teaching people how to use Temporal; they don't need to explain why things do or don't work.

But in any case, I'm familiar with SDK internals and as a reader of this I don't understand what "doesn't respond to things like __name__ that @activity.defn needs" means.

cretz: I can remove those words, no prob, but I do want to explain to people why they can't use the documented way in trio_asyncio which is to use the decorator

dandavison: OK cool, sounds good. Is the issue that people might expect to stack the decorators and that doesn't work? You can avoid using separate sync and async versions of the helper functions like this:

```python
return await trio_asyncio.trio_as_aio(say_hello_in_trio)(name)
```

cretz: > Is the issue that people might expect to stack the decorators and that doesn't work?

Yup, and I just want to mention to them why it doesn't. Otherwise, we should stack the decorators here IMO. Yes, what you have also works, I just chose to use the decorator on that other function instead of this. I will remove the couple of offending words.

cretz: Updated

cretz:
Assuming the other wording is good enough, merging. Can revisit if needed.

PR #167, opened 2025-03-29
Added updatable timer sample.

mfateev:
## What was changed
Added updatable_timer sample based on the [Java one](https://github.com/temporalio/samples-java/tree/main/core/src/main/java/io/temporal/samples/updatabletimer).

## Why?
It demonstrates how to use `workflow.wait_condition` with a timeout. How to update the sleep interval is a FAQ.


updatable_timer/README.md
cretz, line 1: Like our other samples w/ no special install instructions, may want to link to top-level README for "prerequisites"

updatable_timer/workflow.py
```diff
@@ -0,0 +1,35 @@
...
+        )
+        await self.timer.sleep()
+
+    @workflow.signal
+    async def update_wake_up_time(self, wake_up_time: float):
+        # Deals with situation when the signal method is called before the run method.
+        # This happens when a workflow task is executed after a signal is received
+        # or when a workflow is started using the signal-with-start.
+        await workflow.wait_condition(lambda: self.timer is not None)
+        assert self.timer is not None  # for mypy
```
cretz, line 27: We have a neat, fairly new feature for `@workflow.init` that lets you operate on workflow input on the constructor, so can change this to:

```suggestion
@workflow.defn
class Workflow:
    @workflow.init
    def __init__(self, wake_up_time: float) -> None:
        self.timer = UpdatableTimer(
            datetime.fromtimestamp(wake_up_time, tz=timezone.utc)
        )

    @workflow.run
    async def run(self, wake_up_time: float) -> None:
        await self.timer.sleep()

    @workflow.signal
    async def update_wake_up_time(self, wake_up_time: float):
```

(also can remove the `assert self.timer is not None  # for mypy` in the query)

updatable_timer/workflow.py
```diff
@@ -0,0 +1,35 @@
...
+from typing import Optional
+
+from temporalio import workflow
+
+from updatable_timer.updatable_timer_lib import UpdatableTimer
+
+
+@workflow.defn
+class Workflow:
+    def __init__(self):
```
cretz, line 11: Recommend return type annotations everywhere

cretz (commented):
May also want a simple test on this one

updatable_timer/workflow.py
mfateev: Nice! So much clearer!

updatable_timer/workflow.py
mfateev: My Java brain protested this for a constructor :).

cretz (approved):

PR #170, opened 2025-03-31
Migrate to uv

dandavison:
Switch away from poetry to [uv](https://docs.astral.sh/uv).

In the PR's current form, contributors will have to explicitly add samples to this list, at least if they are adding tests or something else that needs to import the sample package. I haven't yet found a way to automatically add top-level directories matching a mattern as packages, but nor have I confirmed that it is impossible. (I believe it's a difference between poetry and uv; the former did this automatically, and the latter perhaps views that as unprincipled.)

```toml
[tool.hatch.build.targets.wheel]
include = ["./**/*.py"]
packages = [
    "activity_worker",
    "bedrock",
    "cloud_export_to_parquet",
    "context_propagation",
    "custom_converter",
    "custom_decorator",
    "dsl",
    "encryption",
    "gevent_async",
    "hello",
    "langchain",
    "message_passing",
    "open_telemetry",
    "patching",
    "polling",
    "prometheus",
    "pydantic_converter",
    "pydantic_converter_v1",
    "pyproject.toml",
    "replay",
    "schedules",
    "sentry",
    "sleep_for_days",
    "tests",
    "trio_async",
    "updatable_timer",
    "worker_specific_task_queues",
    "worker_versioning",
]
```


dandavison:
> Ensure consistency in group naming conventions; confirm whether the identifier should use dashes or underscores across the project.

@copilot see 7bfe1d7d5eb9c1375ee32a7dfc67514c52f4bd01. I believe dashes and underscores can be used interchangeably in group names, although a relevant uv bug was recently fixed: https://github.com/astral-sh/uv/issues/12585. I propose using dashes here: it makes it clear to the reader that what they're looking at are not package names.

README.md
```diff
@@ -1,23 +1,22 @@
 # Temporal Python SDK Samples
 
-This is the set of Python samples for the [Python SDK](https://github.com/temporalio/sdk-python).
+This is a collection of samples showing how to use the [Python SDK](https://github.com/temporalio/sdk-python).
 
 ## Usage
 
 Prerequisites:
 
-* Python >= 3.9
```
cretz, line 9: Can we leave the Python version constraints?

.github/workflows/ci.yml
```diff
@@ -29,10 +30,21 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
-      # Using fixed Poetry version until
-      # https://github.com/python-poetry/poetry/pull/7694 is fixed
-      - run: python -m pip install --upgrade wheel "poetry==1.4.0" poethepoet
-      - run: poetry install --with pydantic_converter --with dsl --with encryption --with trio_async
+      - run: uv tool install poethepoet
+      - run: uv sync --group=dsl --group=encryption --group=trio-async --python ${{ matrix.python }}
+      - name: Check python version
```
cretz, line 35: Not sure we really need this Python version checker code (or the `--python` on the `uv sync`)

dandavison: Oops yes, removed.

README.md
dandavison: We don't want users thinking they have to install a special Python version. uv users should install python using

```
uv python install 3.9
```

dandavison: Happy to clarify somewhere that they should run that command.

cretz: > We don't want users thinking they have to install a special Python version. uv users should install python using

Right, "Python >= 3.9" just means "the software in this repository needs at least Python 3.9", it doesn't say how to install it or that it has to be any specific version. Our SDK and samples repo READMEs need to clarify which versions of Python they work with.

dandavison: How about this?

cretz: LGTM

cretz (approved):

PR #172, opened 2025-04-01
Add a sample that uses a workflow to lock resources

nagl-temporal:
## What was changed
Adds a new sample that demonstrates how to use a workflow to mediate access to a small pool of resources.

## Why?
A customer asked for an example of how to do this kind of thing. We ruled out worker-specific task queues and sessions - those work best when you can run the worker in/near the protected resource. They cannot, for security reasons, which leaves us with:
* non-temporal approaches to resource locking, or
* something like this

## Checklist
1. How was this tested:
Ran the worker and starter per the readme. New unit test passes.

2. Any docs updates needed?
No - it's just a sample.


cretz:
There is a pattern for doing this with signal-with-start. You can see a sample in [.NET](https://github.com/temporalio/samples-dotnet/tree/main/src/Mutex), [Go](https://github.com/temporalio/samples-go/tree/main/mutex), and [TypeScript](https://github.com/temporalio/samples-typescript/tree/main/mutex).

Would totally support such a sample in Python. Needs to be a simple `Lock` class usable in a workflow with similar semantics to https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock but does signal with start and such. Also make sure we have a test or two.

resource_locking/resource_allocator.py
```diff
@@ -0,0 +1,83 @@
...
+            await workflow.wait_condition(lambda: resource is not None, timeout=max_wait_time)
+
+        # During the yield, the calling workflow owns the resource. Note that this is a lock, not a lease! Our
+        # finally block will release the resource if an activity fails. This is why we asserted the lack of
+        # workflow-level timeouts above - the finally block wouldn't run if there was a timeout.
+        try:
+            resource.autorelease = True
+            yield resource
+        finally:
+            if resource.autorelease:
```
nagl-temporal, line 87: Blah. I want there to be a way to tell whether the workflow code is CAN'ing here, but I believe there isn't one.

nagl-temporal:
@cretz LMK what you think about the `async with` abstraction and how it's rough around the edges re: continue-as-new.

resource_locking/resource_allocator.py
```diff
@@ -0,0 +1,106 @@
...
+            start_signal="acquire_resource",
+            start_signal_args=[AcquireRequest(info.workflow_id)],
+        )
+
+    @classmethod
+    @asynccontextmanager
+    async def acquire_resource(
+        cls,
+        *,
+        already_acquired_resource: Optional[AcquiredResource] = None,
```
nagl-temporal, line 49: ⭐

resource_locking/resource_locking_workflow.py
```diff
@@ -0,0 +1,87 @@
...
+# Wait this long for a resource before giving up
+MAX_RESOURCE_WAIT_TIME = timedelta(minutes=5)
+
+
+@workflow.defn(failure_exception_types=[FailWorkflowException])
+class ResourceLockingWorkflow:
+    @workflow.run
+    async def run(self, input: ResourceLockingWorkflowInput):
+        async with ResourceAllocator.acquire_resource(
+            already_acquired_resource=input.already_acquired_resource
```
nagl-temporal, line 61: ⭐ ...I had to have some way of handling the case where my CAN-predecessor already locked the resource. It was either an optional param to `acquire_resource` or branching around the `async with`, which felt much more awkward.

resource_locking/resource_allocator.py
cretz: I think you should always release it if a user is using `async with`. It will be very confusing if I as a user use `async with` and the resource remains when that exits. If we need to let callers keep the resource across continue as new, they should "detach" it (e.g. can offer a helper for this that returns some type) and "reattach" it on next workflow use. Granted I also think including continue as new from the caller side in this sample is a bit confusing, but maybe it's needed.

resource_locking/starter.py
```diff
@@ -0,0 +1,67 @@
...
+        resource_locking_handle = await client.start_workflow(
+            workflow=ResourceLockingWorkflow.run,
+            arg=input,
+            id=f"resource-locking-workflow-{i}",
+            task_queue="default",
+        )
+        resource_locking_handles.append(resource_locking_handle)
+
+    # Add some resources
+    lock_manager_handle = await client.start_workflow(
```
cretz, line 43: Why (maybe) start your one global lock manager after all of its dependents? Why not first?

resource_locking/starter.py
```diff
@@ -0,0 +1,67 @@
...
+        workflow=LockManagerWorkflow.run,
+        arg=LockManagerWorkflowInput(
+            resources={},
+            waiters=[],
+        ),
+        id=LOCK_MANAGER_WORKFLOW_ID,
+        task_queue="default",
+        id_conflict_policy=WorkflowIDConflictPolicy.USE_EXISTING,
+        start_signal="add_resources",
+        start_signal_args=[["resource_a", "resource_b"]],
```
cretz, line 53: Consider also accepting the resources in the workflow input so you don't have to have this extra event/action.

nagl-temporal:
Thank you - I like the new names. I'll take most of the comments. I want to leave CAN in, because that's the only thing that makes this tricky - sample feels incomplete without it. However, if I add `.lock()` and `.unlock()` and use those, the awkwardness goes away.

resource_locking/lock_manager_workflow.py
```diff
@@ -0,0 +1,130 @@
...
+
+        requester = workflow.get_external_workflow_handle(internal_request.workflow_id)
+        await requester.signal(
+            "assign_resource",
+            AcquireResponse(
+                release_signal_name=internal_request.release_signal, resource=resource
+            ),
+        )
+
+    @workflow.signal(dynamic=True)
```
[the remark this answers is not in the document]
nagl-temporal: The answer to "why" is that a dynamic handler lets me log spurious signals. But thanks for pressing on it... that's not enough of a reason to use one.

resource_locking/lock_manager_workflow.py
```diff
@@ -0,0 +1,130 @@
...
+        del self.release_signal_to_resource[signal_name]
+
+        # If there are queued requests, assign the resource to the next one
+        if len(self.waiters) > 0:
+            next_holder = self.waiters.pop(0)
+            await self.assign_resource(resource, next_holder)
+
+    @workflow.query
+    def get_current_holders(self) -> dict[str, Optional[InternalAcquireRequest]]:
+        return {k: v if v else None for k, v in self.resources.items()}
```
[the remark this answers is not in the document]
nagl-temporal: Doh. This transform used to do more.

resource_locking/resource_allocator.py
```diff
@@ -0,0 +1,106 @@
...
+class ResourceAllocator:
+    def __init__(self, client: Client):
+        self.client = client
+
+    @activity.defn
+    async def send_acquire_signal(self) -> None:
+        info = activity.info()
+
+        # This will start and signal the workflow if it isn't running, otherwise it will signal the current run.
+        await self.client.start_workflow(
```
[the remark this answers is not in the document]
nagl-temporal: That's what I had before (specifically so that I didn't have to drop into an activity). I switched to align with the other samples per request. But I think we've drifted away from that - I'll put the signal back.

resource_locking/shared.py
```diff
@@ -0,0 +1,19 @@
+from dataclasses import dataclass, field
+
+LOCK_MANAGER_WORKFLOW_ID = "lock_manager"
```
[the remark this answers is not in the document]
nagl-temporal: Agreed - this was just sample simplification. It's easy to do better though, you're right.

resource_locking/resource_allocator.py
```diff
@@ -0,0 +1,106 @@
...
+    AcquiredResource,
+    AcquireRequest,
+    AcquireResponse,
+)
+
+
+# Use this class in workflow code that that needs to run on locked resources.
+class ResourceAllocator:
+    def __init__(self, client: Client):
+        self.client = client
```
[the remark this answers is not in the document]
nagl-temporal: Thanks, that makes sense. Backing away from signal-with-start makes it all just workflow code, so there ended up being nothing to separate.

resource_locking/resource_allocator.py
```diff
@@ -0,0 +1,106 @@
...
+from resource_locking.shared import (
+    LOCK_MANAGER_WORKFLOW_ID,
+    AcquiredResource,
+    AcquireRequest,
+    AcquireResponse,
+)
+
+
+# Use this class in workflow code that that needs to run on locked resources.
+class ResourceAllocator:
```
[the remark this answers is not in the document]
nagl-temporal: It became `ResourcePoolClient`.

resource_locking/resource_allocator.py
```diff
@@ -0,0 +1,106 @@
...
+                workflow.set_signal_handler("assign_resource", None)
+                nonlocal resource
+                resource = AcquiredResource(
+                    resource=input.resource,
+                    release_signal_name=input.release_signal_name,
+                )
+
+            workflow.set_signal_handler("assign_resource", assign_resource)
+
+            await workflow.start_activity(
```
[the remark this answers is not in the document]
nagl-temporal: The activity went away.

resource_locking/resource_allocator.py
```diff
@@ -0,0 +1,106 @@
...
+                ResourceAllocator.send_acquire_signal,  # type: ignore[arg-type]
+                start_to_close_timeout=timedelta(seconds=10),
```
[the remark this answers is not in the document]
nagl-temporal: The activity went away.

resource_locking/resource_allocator.py
```diff
@@ -0,0 +1,106 @@
...
+            workflow=LockManagerWorkflow.run,
+            arg=LockManagerWorkflowInput(
+                resources={},
+                waiters=[],
+            ),
+            id=LOCK_MANAGER_WORKFLOW_ID,
+            task_queue="default",
+            id_conflict_policy=WorkflowIDConflictPolicy.USE_EXISTING,
+            start_signal="acquire_resource",
+            start_signal_args=[AcquireRequest(info.workflow_id)],
```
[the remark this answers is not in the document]
nagl-temporal: I do think it's important to keep CAN in. Without CAN, the sample can't demonstrate how to detach/reattach so it seems incomplete to me.

resource_locking/starter.py
nagl-temporal: It was to exercise signal-with-start in the dependents, but the need for that is gone now.

resource_pool/starter.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # Initialize the resource pool
+    resource_pool_handle = await client.start_workflow(
+        workflow=ResourcePoolWorkflow.run,
+        arg=ResourcePoolWorkflowInput(
+            resources={"resource_a": None, "resource_b": None},
+            waiters=[],
+        ),
+        id=RESOURCE_POOL_WORKFLOW_ID,
+        task_queue="default",
```
cretz, line 30:
```suggestion
        task_queue="resource_pool-task-queue",
```

Here and everywhere just to avoid clash with other samples

resource_pool/resource_pool_client.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+# Use this class in workflow code that that needs to run on locked resources.
+class ResourcePoolClient:
+    def __init__(self, pool_workflow_id: str) -> None:
+        self.pool_workflow_id = pool_workflow_id
+        self.acquired_resources: list[AcquiredResource] = []
+
+    async def send_acquire_signal(self) -> None:
+        handle = workflow.get_external_workflow_handle(self.pool_workflow_id)
```
cretz, line 22:
```suggestion
        handle = workflow.get_external_workflow_handle_for(ResourcePoolWorkflow.run, self.pool_workflow_id)
```

To be more type safe

resource_pool/resource_pool_client.py
```diff
@@ -0,0 +1,97 @@
...
+        await handle.signal(
+            "acquire_resource", AcquireRequest(workflow.info().workflow_id)
+        )
```
cretz, line 25:
```suggestion
        await handle.signal(
            ResourcePoolWorkflow.acquire_resource, AcquireRequest(workflow.info().workflow_id)
        )
```

To be more type safe. Same everywhere

resource_pool/resource_pool_client.py
```diff
@@ -0,0 +1,97 @@
...
+        # finally block will release the resource if an activity fails. This is why we asserted the lack of
+        # workflow-level timeouts above - the finally block wouldn't run if there was a timeout.
+        try:
+            yield resource
+        finally:
+            if not resource.detached:
+                await self.send_release_signal(resource)
+
+
+def warn_when_workflow_has_timeouts() -> None:
```
cretz, line 83: I'd toss underscores in front of these to make it clear they are only for use in this module (or just inline the single-use function). Same for the other top-level function below.

resource_pool/resource_pool_client.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @asynccontextmanager
+    async def acquire_resource(
+        self,
+        *,
+        reattach: Optional[DetachedResource] = None,
+        max_wait_time: timedelta = timedelta(minutes=5),
+    ) -> AsyncGenerator[AcquiredResource, None]:
+        warn_when_workflow_has_timeouts()
+        self.lazy_register_signal_handler()
```
cretz, line 56: Can do this non-lazy on instantiation too

resource_pool/resource_pool_client.py
```diff
@@ -0,0 +1,97 @@
...
+        await handle.signal(
+            "release_resource",
+            AcquireResponse(
+                resource=acquired_resource.resource,
+                release_key=acquired_resource.release_key,
+            ),
+        )
+
+    def lazy_register_signal_handler(self) -> None:
+        if workflow.get_signal_handler("assign_resource") is None:
```
cretz, line 38: This name prevents use of multiple resource pools clients in a single workflow. If that's ok for this sample, might as well change the constructor of this client to have the workflow ID defaulted to the known, hardcoded, shared ID (or just remove it as a parameter altogether)

resource_pool/resource_pool_client.py
cretz, line 1: Pedantic, but could move this, the pool workflow, and the shared stuff into a separate dir/module to clarify it is to be seen as a reusable lib

resource_pool/resource_pool_workflow.py
```diff
@@ -0,0 +1,141 @@
...
+    @workflow.signal
+    async def acquire_resource(self, request: AcquireRequest) -> None:
+        self.waiters.append(
+            InternalAcquireRequest(workflow_id=request.workflow_id, release_signal=None)
+        )
+        workflow.logger.info(
+            f"workflow_id={request.workflow_id} is waiting for a resource"
+        )
+
+    @workflow.signal()
```
cretz, line 54:
```suggestion
    @workflow.signal
```

resource_pool/resource_pool_workflow.py
```diff
@@ -0,0 +1,141 @@
...
+                self.release_key_to_resource[holder.release_signal] = resource
+
+    @workflow.signal
+    async def add_resources(self, resources: list[str]) -> None:
+        for resource in resources:
+            if resource in self.resources:
+                workflow.logger.warning(
+                    f"Ignoring attempt to add already-existing resource: {resource}"
+                )
+                continue
```
cretz, line 41:
```suggestion
```

resource_pool/resource_pool_workflow.py
```diff
@@ -0,0 +1,141 @@
...
+    ) -> None:
+        self.resources[resource] = internal_request
+        workflow.logger.info(
+            f"workflow_id={internal_request.workflow_id} acquired resource {resource}"
+        )
+        internal_request.release_signal = str(workflow.uuid4())
+        self.release_key_to_resource[internal_request.release_signal] = resource
+
+        requester = workflow.get_external_workflow_handle(internal_request.workflow_id)
+        await requester.signal(
```
cretz, line 91: Note that if a waiting workflow is no longer running for any reason, this signal will fail causing the this whole pool to fail. Not a big deal for a sample, but technically catching exception for inability to signal external workflow may be clearer.

tests/resource_locking/__init__.py
cretz, line 1: Dir should be `tests/resource_pool`

tests/resource_locking/workflow_test.py
```diff
@@ -0,0 +1,111 @@
...
+    ResourcePoolWorkflowInput,
+)
+from resource_pool.resource_user_workflow import (
+    ResourceUserWorkflow,
+    ResourceUserWorkflowInput,
+    UseResourceActivityInput,
+)
+from resource_pool.shared import RESOURCE_POOL_WORKFLOW_ID
+
+TASK_QUEUE = "default"
```
cretz, line 21: Can this be changed to avoid test clashing? Even a UUID would be fine.

cretz (commented):
Looks great, only a few minor things

resource_pool/resource_pool_client.py
nagl-temporal: Out of curiosity, this can't help the act of sending the signal, right? It's just there as a best practice for copy/pasters?

resource_pool/resource_pool_client.py
nagl-temporal: Almost everywhere - the pool workflow should signal the requester in an un-typesafe way (since it can't know what kind of workflow the requester is).

resource_pool/resource_pool_client.py
cretz: It actually can in a neat way. When you use `handle.signal(ResourcePoolWorkflow.acquire_resource` below, that does a type check not only on the argument type, but also that that signal is even on the same class. So if you had `handle.signal(SomeOtherWorkflow.acquire_resource` it would fail type check since the generic on the external signal handle is the "self type".

resource_pool/resource_pool_client.py
nagl-temporal: Good callout. I prefer to include the workflow id in the signal name.

resource_pool/resource_pool_client.py
nagl-temporal: Oh that's neat. I didn't look deeply enough to find the wiring that enforces what you described.

resource_pool/resource_pool_workflow.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+
+        requester = workflow.get_external_workflow_handle(internal_request.workflow_id)
+        try:
+            release_signal = str(workflow.uuid4())
+            await requester.signal(
+                f"assign_resource_{workflow.info().workflow_id}",
+                AcquireResponse(release_key=release_signal, resource=resource),
+            )
+
```
nagl-temporal, line 94: There's no race here, right? Imagine the release signal arrives immediately. The fix is obvious if there is a race, but the code smells nicer this way.

Why I think there's no race: I believe that this handler must run to completion before the handler for the release signal can start.

resource_pool/resource_pool_workflow.py
cretz, line 1: I was thinking this file would be inside the module too since it's also part of this "pool" library that users use instead of part of user code

resource_pool/resource_pool_workflow.py
cretz: Correct, there is no race here because this is only called in one place, serially, in the primary loop. However, this can show the dangers of overly extracting/modularizing single-use methods - it can be hard to see the constraints they expect of the callers. Can technically add a "Not safe for concurrent use" docstring if concerned.

resource_pool/pool_client/resource_pool_client.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+# Use this class in workflow code that that needs to run on locked resources.
+class ResourcePoolClient:
+    def __init__(self, pool_workflow_id: str) -> None:
+        self.pool_workflow_id = pool_workflow_id
+        self.acquired_resources: list[AcquiredResource] = []
+        self.register_signal_handler()
+
+    def register_signal_handler(self) -> None:
```
cretz, line 23:
```suggestion
```

Really gaining anything be making these few lines a separate method?

resource_pool/pool_client/resource_pool_client.py
```diff
@@ -0,0 +1,103 @@
...
+        signal_name = f"assign_resource_{self.pool_workflow_id}"
+        if workflow.get_signal_handler(signal_name) is None:
+            workflow.set_signal_handler(signal_name, self.assign_resource)
+        else:
+            raise RuntimeError(
+                f"{signal_name} already registered - if you use multiple ResourcePoolClients within the "
+                f"same workflow, they must use different pool_workflow_ids"
+            )
+
+    async def send_acquire_signal(self) -> None:
```
cretz, line 33: Should prefix all methods in this class that you don't expect an end user to call with `_`

cretz (approved):
Only minor, non-blocking things, LGTM. Merge at will or let me know when done and I can.

PR #173, opened 2025-04-02
Convert hello activities to sync

GSmithApps:
## What was changed
Changed the activities in `hello` to be synchronous

## Why?
To conform to our recommendation of using sync activities unless developers are *very* sure that their async activities aren't blocking

## Checklist

1. How was this tested:
ran all the samples and their behavior was unchanged

2. Any docs updates needed?
no need

## Notes to reviewers
This is a proposal. Please feel free to edit however you see fit. Thank you!


cretz (approved):
Looks great, thanks! We should do this to other samples too, but this is a good start. @dandavison - thoughts?

hello/hello_async_activity_completion.py
```diff
@@ -27,29 +30,28 @@ async def compose_greeting(self, input: ComposeGreetingInput) -> str:
         # So we store the tasks ourselves.
         # See https://docs.python.org/3/library/asyncio-task.html#creating-tasks,
         # https://bugs.python.org/issue21163 and others.
-        _ = asyncio.create_task(
-            self.complete_greeting(activity.info().task_token, input)
-        )
+        Thread(
+            target=self.complete_greeting,
+            args=(activity.info().task_token, input),
+        ).start()
```
dandavison, line 36: Is this definitely our recommendation or is this a case where it feels more natural for the user to use an `asyncio` task than to have to explicitly create a thread, seeing as they use asyncio elsewhere with the SDK?

hello/hello_async_activity_completion.py
```diff
@@ -27,29 +30,28 @@ async def compose_greeting(self, input: ComposeGreetingInput) -> str:
...
         for _ in range(0, 3):
             print("Waiting one second...")
-            await handle.heartbeat()
-            await asyncio.sleep(1)
+            asyncio.run(handle.heartbeat())
+            time.sleep(1)
 
         # Complete using the handle
-        await handle.complete(f"{input.greeting}, {input.name}!")
+        asyncio.run(handle.complete(f"{input.greeting}, {input.name}!"))
```
dandavison, line 54: Using `asyncio.run` twice in a function in a expositional example surprises me (but I'm new to the sync vs async activity recommendation discussion)

GSmithApps:
Whoa whoa I made a git mistake -- please don't merge

hello/hello_async_activity_completion.py
GSmithApps: I'm okay either way on this! Up to you folks. I went ahead and removed the comments though ✅

hello/hello_async_activity_completion.py
GSmithApps: Thanks so much for catching this! Should be fixed 👍

GSmithApps:
Ready if you guys are 🚀

hello/hello_async_activity_completion.py
```diff
@@ -14,42 +17,39 @@ class ComposeGreetingInput:
...
         # Let's wait three seconds, heartbeating each second. Note, heartbeating
         # during async activity completion is done via the client directly. It
         # is often important to heartbeat so the server can know when an
         # activity has crashed.
         handle = self.client.get_async_activity_handle(task_token=task_token)
         for _ in range(0, 3):
             print("Waiting one second...")
-            await handle.heartbeat()
-            await asyncio.sleep(1)
+            asyncio.run_coroutine_threadsafe(handle.heartbeat(), self.loop)
```
cretz, line 46:
```suggestion
            asyncio.run_coroutine_threadsafe(handle.heartbeat(), self.loop).result()
```

Here and on the one below so you actually wait on the future. Also we recognize a lack of tests on these hello samples, so can you at least confirm when manually running them they work as expected?

cretz (approved):
LGTM, but wait for @dandavison's approval

hello/hello_async_activity_completion.py
GSmithApps: Thank you!

Yeah i put the results in [this ticket](https://temporalio.atlassian.net/jira/software/c/projects/DSE/boards/330?selectedIssue=DSE-73). Unfortunately it's internal. I could attach them here if we want

hello/hello_activity.py
```diff
@@ -50,6 +51,10 @@ async def main():
         task_queue="hello-activity-task-queue",
         workflows=[GreetingWorkflow],
         activities=[compose_greeting],
+        # Non-async activities require an executor;
+        # a thread pool executor is recommended.
+        # This same thread pool could be passed to multiple workers if desired.
+        activity_executor=ThreadPoolExecutor(5),
```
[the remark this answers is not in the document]
GSmithApps: I don't think i got those warnings. See output in [this ticket](https://temporalio.atlassian.net/jira/software/c/projects/DSE/boards/330?selectedIssue=DSE-73). I can double check though

Or I can just go ahead and up it to 100 anyway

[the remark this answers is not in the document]
cretz: Interesting, I wonder why they aren't showing

cretz (approved):
LGTM, @dandavison - this look good to you?

dandavison:
I'm not sure I agree with this. Activities will typically do network I/O, and the Python ecosystem has been moving more and more in the direction of using `async`/`await` to do I/O for years. I understand that some users have accidentally blocked event loops by using thread-blocking libraries, but I am not yet convinced that that merits switching everything away from `async def` and making it look like the experience of using Temporal doesn't involve using `async`/`await` in activities.

GSmithApps:
This is being discussed off-PR

hello/hello_async_activity_completion.py
```diff
@@ -14,42 +17,39 @@ class ComposeGreetingInput:
...
         # Let's wait three seconds, heartbeating each second. Note, heartbeating
         # during async activity completion is done via the client directly. It
         # is often important to heartbeat so the server can know when an
         # activity has crashed.
         handle = self.client.get_async_activity_handle(task_token=task_token)
         for _ in range(0, 3):
             print("Waiting one second...")
-            await handle.heartbeat()
-            await asyncio.sleep(1)
+            asyncio.run_coroutine_threadsafe(handle.heartbeat(), self.loop).result()
```
dandavison, line 46: Would this sample be better left using `asyncio`? It's using obscure `asyncio` APIs that most people have never heard of, and has to pass in an event loop. It feels like it's working against the grain of the library, and it needs some explanation of what problem this sample's helping people solve, and changing it to sync has made even harder to understand I fear.

dandavison (approved):
We agreed offline that this was the way to go (apart from anything else, users of async I/O libraries will likely know that they need to use `async/await`, whereas users of traditional thread-blocking libraries may not realize they have to use an executor).

My remaining suggestion is that one async activity completion sample that seemed to go off the deep end. Looks to me like we should maybe keep that one as is until we've made it more intuitive to do that in a non-`async` way. It even has `async` in its name already!

PR #174, opened 2025-04-02
Nexus samples

dandavison:
Rendered: https://github.com/temporalio/samples-python/tree/nexus/hello_nexus


hello_nexus/basic/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+```
+temporal operator namespace create --namespace my-target-namespace
+temporal operator namespace create --namespace my-caller-namespace
+
+temporal operator nexus endpoint create \
+  --name my-nexus-endpoint \
+  --target-namespace my-target-namespace \
+  --target-task-queue my-target-task-queue \
+  --description-file ./hello_nexus/basic/service_description.md
```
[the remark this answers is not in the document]
dandavison: Done

hello_nexus/basic/caller/__init__.py
```diff
@@ -0,0 +1 @@
+from . import app as app
```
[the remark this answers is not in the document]
dandavison: Nope, thanks

hello_nexus/basic/README.md
```diff
@@ -0,0 +1,37 @@
...
+### Sample directory structure
+
+- [service.py](./service.py) - shared Nexus service definition
+- [caller](./caller) - a caller workflow that executes Nexus operations, together with a worker and starter code
+- [handler](./handler) - Nexus operation handlers, together with a workflow used by one of the Nexus operations, and a worker that polls for both workflow and Nexus tasks.
+
+
+### Instructions
+
+Start a Temporal server.
```
cretz, line 13: Recommend linking to primary README for this or somewhere that explains this

hello_nexus/basic/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Run the following:
+
+```
+temporal operator namespace create --namespace my-handler-namespace
+temporal operator namespace create --namespace my-caller-namespace
```
cretz, line 19: We support providing namespace at `start-dev` time, that may be easier for users

hello_nexus/basic/README.md
cretz, line 1: This file assumes you are `cd`'d into the parent dir and not this dir?

hello_nexus/basic/caller/app.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Optional
+
+from temporalio.client import Client
+from temporalio.worker import UnsandboxedWorkflowRunner, Worker
+
+from hello_nexus.basic.caller.workflows import CallerWorkflow
+from hello_nexus.basic.service import MyOutput
+
+NAMESPACE = "my-caller-namespace"
+TASK_QUEUE = "my-caller-task-queue"
```
cretz, line 12: Usually in samples, all task queues and workflow IDs are prefixed with something about the sample so they don't clash with other samples and are clear what they are for (and in this unique multi-namespace sample, I'd apply the same logic to namespace names and endpoint names too)

hello_nexus/basic/caller/app.py
```diff
@@ -0,0 +1,46 @@
...
+        namespace=NAMESPACE,
+    )
+
+    async with Worker(
+        client,
+        task_queue=TASK_QUEUE,
+        workflows=[CallerWorkflow],
+        # TODO(dan): isinstance(op, nexusrpc.contract.Operation) is failing under the
+        # sandbox in temporalio/worker/_interceptor.py
+        workflow_runner=UnsandboxedWorkflowRunner(),
```
cretz, line 29: I think the Temporal SDK needs to auto-pass-through the entire `nexusrpc` module (which will include children) by default same as we do with `temporalio`

hello_nexus/basic/caller/workflows.py
cretz, line 1: While by request I did ignore the workflow caller impl details on Temporal SDK PR, a quick review of this from a user POV LGTM

hello_nexus/basic/handler/service_handler.py
```diff
@@ -0,0 +1,86 @@
...
+    # that unlike the workflow run operation above, in this case the `start` method
+    # returns the final operation result.
+    #
+    # Here it is implemented using SyncOperationHandler.from_callable.
+    # See service_handler_with_operation_handler_classes.py for an alternative style
+    # involving subclassing SyncOperationHandler and overriding the start method.
+    #
+    # Sync operations are free to make arbitrary network calls, or perform CPU-bound
+    # computations. Total execution duration must not exceed 10s.
+    @operation_handler
```
cretz, line 78: I recall a `@sync_operation_handler` of sorts that would have saved lines here. Did we get rid of that? Do we expect users writing very simple synchronous Nexus ops to have to make nested functions or classes?

hello_nexus/without_service_definition/app.py
```diff
@@ -0,0 +1,117 @@
...
+@workflow.defn
+class HandlerWorkflow:
+    @workflow.run
+    async def run(self, message: str) -> str:
+        return f"Hello {message} from workflow run operation!"
+
+
+# Here we define a nexus service by providing a service handler implementation without a
+# service contract. This nexus service has one operation.
+@service_handler
```
cretz, line 46: I think if we're willing to have multiple service impl approaches in the basic sample for operation handlers, we can toss this in there too. I think the basic sample having 3 different services implemented different ways is more reasonable than 1 sample with 2 different service approaches and another with a third.

Alternatively, if we keep this separate, I think we should consider breaking files out similarly to the basic sample.

tests/hello_nexus/helpers.py
```diff
@@ -0,0 +1,41 @@
...
+import temporalio.api.nexus.v1
+import temporalio.api.operatorservice
+import temporalio.api.operatorservice.v1
+import temporalio.nexus
+import temporalio.nexus.handler
+from temporalio.client import Client
+
+
+# TODO: copied from sdk-python tests/helpers/nexus
+async def create_nexus_endpoint(
```
cretz, line 15: So this is fairly bad dx I think for testing Nexus from a user POV. Can we open an issue in CLI (but may need to be in server first) to allow Nexus endpoints to be created as part of `server start-dev` similar to what we do with namespaces?

tests/hello_nexus/basic_test.py
cretz, line 1: A common use case for users doing tests is to mock the side-effecting calls made by the workflow. So in our samples tests for basic activities we show how to mock the activity in the test. We also have this front and center at https://docs.temporal.io/develop/python/testing-suite#mock-activities because it's so often requested.

Can we show a user how to test a simple workflow but w/ Nexus operation mocked? If we need to make changes to the SDK to make this easier, we should consider it.

hello_nexus/basic/handler/service_handler.py
```diff
@@ -0,0 +1,69 @@
...
+    # This is a nexus operation that is backed by a Temporal workflow. The start method
+    # starts a workflow, and returns a nexus operation token that the handler can use to
+    # obtain a workflow handle (for example if a cancel request is subsequently sent by
+    # the caller). The Temporal server takes care of delivering the workflow result to the
+    # calling workflow. The task queue defaults to the task queue being used by the Nexus
+    # worker.
+    @temporalio.nexus.handler.workflow_run_operation_handler
+    async def my_workflow_run_operation(
+        self, ctx: StartOperationContext, input: MyInput
+    ) -> WorkflowHandle[WorkflowStartedByNexusOperation, MyOutput]:
```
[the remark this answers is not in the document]
dandavison: Thanks. The current design is it returns `nexus.WorkflowHandle`. I've confirmed that a type-safe `fetch_result` implementation is impossible, via run-time access to the generic type params.

hello_nexus/basic/caller/workflows.py
```diff
@@ -0,0 +1,35 @@
+from temporalio import workflow
+from temporalio.workflow import NexusClient
+
+from hello_nexus.basic.service import MyInput, MyNexusService, MyOutput
```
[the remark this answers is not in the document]
dandavison: Thanks, done.

hello_nexus/basic/handler/service_handler.py
```diff
@@ -0,0 +1,86 @@
...
+            ctx: StartOperationContext, input: MyInput
+        ) -> WorkflowOperationToken[MyOutput]:
+            # You could use self.connected_db_client here.
+            return await start_workflow(
+                WorkflowStartedByNexusOperation.run,
+                input,
+                id=str(uuid.uuid4()),
+            )
+
+        return WorkflowRunOperationHandler.from_callable(start)
```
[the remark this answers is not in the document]
dandavison: This design has been superseded

hello_nexus/basic/README.md
dandavison: Done thanks

hello_nexus/basic/handler/db_client.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import annotations
+
+
+class MyDBClient:
```
bergundy, line 4: I would be careful encouraging users to make DB calls in sync handlers. Sync handlers should typically be calling into Temporal APIs. The reason being that the caller controls retrying nexus start operation requests and if there's a chance of failure, we recommend deferring the action to be done in an activity (standalone activities will complete the picture here).
I think there is value in showing dependency injection in the sample, just warn users in comments about the retry caveat.

hello_nexus/basic/handler/service_handler.py
```diff
@@ -0,0 +1,58 @@
...
+@service_handler(service=MyNexusService)
+class MyNexusServiceHandler:
+    # You can create an __init__ method accepting what is needed by your operation
+    # handlers to handle requests. You typically instantiate your service handler class
+    # when starting your worker. See hello_nexus/basic/handler/worker.py.
+    def __init__(self, connected_db_client: MyDBClient):
+        # `connected_db_client` is intended as an example of something that might be
+        # required by your operation handlers when handling requests, but is only
+        # available at worker-start time.
+        self.connected_db_client = connected_db_client
```
bergundy, line 27: Maybe show how this may be used in the sample, perhaps in the sync handler...

hello_nexus/README.md
```diff
@@ -0,0 +1,15 @@
...
+Workflows, or execute arbitrary code.
+
+Learn more at [temporal.io/nexus](https://temporal.io/nexus).
+
+The samples in this directory form an introduction to Nexus. 
+
+### Samples
+
+- [basic](./basic) - Nexus service definition, operation handlers, and calling workflows.
+- [without_service_definition](./without_service_definition) - A Nexus service implementation without a service definition
```
bergundy, line 15: Not sure we want to show this in the samples, we should recommend using a service contract when calling across namespaces IMHO. The contract-less calls would have more value when we support same-team, direct task queue operation invocations.

bergundy (commented):
Overall LGTM. Didn't have much to add.

hello_nexus/handler/workflows.py
```diff
@@ -0,0 +1,12 @@
+from temporalio import workflow
+
+with workflow.unsafe.imports_passed_through():
```
bergundy, line 3: Can we make this not required? Seems like everything in `nexusrpc` should be safe to import (except for the `handler` submodule)

bergundy (approved):

PR #177, opened 2025-04-18
added custom metric sample

GSmithApps:
## What was changed
Added a custom metric sample

## Why?
Help folks with questions about custom metrics

## Checklist

1. How was this tested:
Checked the metric output against the existing metric

2. Any docs updates needed?
Updated main python samples readme


custom_metric/activity_worker.py
```diff
@@ -0,0 +1,71 @@
...
+        meter = activity.metric_meter()
+        histogram = meter.create_histogram_timedelta(
+            "custom_activity_schedule_to_start_latency",
+            description="Time between activity scheduling and start",
+            unit="duration",
+        )
+        histogram.record(
+            schedule_to_start, {"workflow_type": activity.info().workflow_type}
+        )
+        return await self.next.execute_activity(input)
```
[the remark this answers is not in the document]
GSmithApps: This is set in the base class. Not sure if we want to heed this suggestion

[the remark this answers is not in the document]
cretz: Nope, it's wrong (SDK team doesn't traditionally use these AI tools in PRs, not that we're against it or anything)

custom_metric/README.md
```diff
@@ -0,0 +1,33 @@
+# Custom Metric
+
+1. Run the server with `temporal server start-dev`
+2. Run the client with `uv run custom_metric/client.py`
+3. Run the workflow worker with `uv run custom_metric/workflow_worker.py`
+4. Run the activity worker with `uv run custom_metric/activity_worker.py`
```
cretz, line 6: Why separate workflow worker and activity worker instead of just having one worker?

custom_metric/README.md
```diff
@@ -0,0 +1,33 @@
+# Custom Metric
```
cretz, line 1: Can you give one or two sentences here on the purpose and/or what the sample will demonstrate?

custom_metric/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+1. Run the server with `temporal server start-dev`
```
cretz, line 3: Have the sample refer to the top-level README for prerequisites like Python version, starting local server, etc. We try to keep this in a common place.

custom_metric/activity_worker.py
```diff
@@ -0,0 +1,71 @@
...
+from temporalio.runtime import PrometheusConfig, Runtime, TelemetryConfig
+from temporalio.worker import (
+    ActivityInboundInterceptor,
+    ExecuteActivityInput,
+    Interceptor,
+    Worker,
+)
+
+
+class SimpleWorkerInterceptor(Interceptor):
```
cretz, line 15: This sample is not just demonstrating custom metric, which can be done inside the activity, it's demonstrating interceptors. May be a little confusing to someone just wanting to understand the custom metric part.

custom_metric/README.md
cretz, line 1: Can you add a test or two? (I know not all samples have tests, something we regret, but adding them helps us catch if they start breaking)

custom_metric/README.md
GSmithApps: Because it's a histogram metric and I want to show the histogram bins, so I separated the workers to have a little delay in there. I suppose it's not 100% necessary

custom_metric/activity_worker.py
GSmithApps: Thanks for the suggestion! I implemented it because what you mentioned off-PR, I thought you suggested to do this with an interceptor. Am I missing something?

custom_metric/README.md
cretz: Yeah it is a bit confusing to do and was not obvious that was the purpose, so people may read this and think "oh for custom metrics I have to split the workers". If you wanted to show real-world schedule to start latency, have a worker with max concurrent activities as 1, a workflow that starts 3 activities on it concurrently, and each activity sleep a tad.

custom_metric/activity_worker.py
cretz: For _this_ specific custom metric that makes sense, but people can create custom metrics without interceptors. I think it's fine to stay, just need to clarify in README that the purpose of this sample is to demonstrate a custom schedule to start metric _via interceptor_, as opposed to just general custom metric.

custom_metric/README.md
GSmithApps: Added!

custom_metric/activity_worker.py
GSmithApps: I added a little blurb. I hope it's along the lines you're thinking

GSmithApps:
@cretz thanks for the review! I think all your comments have been acknowledged and fixed

custom_metric/worker.py
```diff
@@ -0,0 +1,97 @@
...
+)
+
+
+@activity.defn
+def print_message():
+    print("In the activity.")
+    time.sleep(1)
+
+
+@workflow.defn
```
cretz, line 23: Pedantic, but can you move the workflow to its own file (and pass through the import used to reference the activity)? We have issues currently where users are reading our samples and thinking their workflows should be in the same file with other things.

tests/custom_metric/workflow_test.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+async def test_sleep_for_days_workflow():
+
+    @activity.defn(name="print_message")
+    async def print_message_mock():
+        global activity_counter
+        activity_counter += 1
+
+    async with await WorkflowEnvironment.start_time_skipping() as env:
```
cretz, line 21:
```suggestion
async def test_sleep_for_days_workflow(client: Client):

    @activity.defn(name="print_message")
    async def print_message_mock():
        global activity_counter
        activity_counter += 1

```

I think we should encourage use of the shared dev server and only encourage use of time skipping server when it's needed (other tests may be unnecessarily using time skipping test server, but they shouldn't IMO)

custom_metric/README.md
```diff
@@ -0,0 +1,36 @@
+# Custom Metric
+
+This sample deminstrates two things: (1) how to make a custom metric, and (2) how to use an interceptor.
+The custom metric in this sample is an activity schedule-to-start-latency with a workflow type tag.
+
+Please see the top-level README for prerequisites such as Python, uv, starting the local temporal development server, etc.
+
+1. Run the worker with `uv run custom_metric/worker.py`
+2. Request execution of the workflow with `temporal workflow start --type ExecuteActivityWorkflow --task-queue custom-metric-task-queue`
```
cretz, line 9: Usually we prefer a `starter.py` over encouraging CLI use, but not a big deal

custom_metric/README.md
```diff
@@ -0,0 +1,36 @@
+# Custom Metric
+
+This sample deminstrates two things: (1) how to make a custom metric, and (2) how to use an interceptor.
+The custom metric in this sample is an activity schedule-to-start-latency with a workflow type tag.
+
+Please see the top-level README for prerequisites such as Python, uv, starting the local temporal development server, etc.
```
cretz, line 6:
```suggestion
Please see the top-level [README](../README.md) for prerequisites such as Python, uv, starting the local temporal development server, etc.
```

cretz (commented):
Just minor stuff. Only thing really needed I think is to give the workflow its own file

tests/custom_metric/workflow_test.py
```diff
@@ -0,0 +1,32 @@
...
+activity_counter = 0
+
+
+async def test_custom_metric_workflow(client: Client):
+    @activity.defn(name="print_message")
+    async def print_message_mock():
+        global activity_counter
+        activity_counter += 1
+
+        async with Worker(
```
cretz, line 20: I don't think you want the worker _inside_ the activity 😁

GSmithApps: 🤦

cretz (approved):
Looks great, thanks!

PR #183, opened 2025-04-30
fix: periodic polling sample

### LINKED ISSUE #182: [Bug] continue_as_new prematurely exits retry loop in "Periodic Polling of a Sequence of Activities" example
### What are you really trying to do?

I'm trying to implement the pattern described in the "Periodic Polling of a Sequence of Activities" example, where a workflow periodically polls by running an activity in a loop with delays and limited retries. The goal is to retry the activity up to 10 times (with 1-second intervals) before giving up, and only use `continue_as_new` after exhausting those attempts.

### Describe the bug

The `workflow.continue_as_new(name)` statement is placed inside the `for` loop. This causes the workflow to continue as new after **just one failed attempt**, which resets the retry counter and prevents the loop from reaching the intended 10 retries. As a result, the activity is never actually retried 10 times within a single workflow execution.

This breaks the expected retry behavior and causes a new workflow execution to be spawned each time there's a failure, which is not the intended design.

### Minimal Reproduction

```diff
@workflow.defn
class ChildWorkflow:
    @workflow.run
    async def run(self, name: str) -> str:
        for i in range(10):
            try:
                return await workflow.execute_activity(
                    compose_greeting,
                    ComposeGreetingInput("Hello", name),
                    start_to_close_timeout=timedelta(seconds=4),
                    retry_policy=RetryPolicy(
                        maximum_attempts=1,
                    ),
                )
            except ActivityError:
                workflow.logger.error("Activity failed, retrying in 1 seconds")
            await asyncio.sleep(1)
-            workflow.continue_as_new(name)  # This breaks the retry loop

        raise Exception("Polling failed after all attempts")
```

Nomango:
## What was changed

Fixed the retry loop in the "Periodic Polling of a Sequence of Activities" example.

## Why?
Previously, `continue_as_new` was inside the retry loop, causing the workflow to restart after just **one failed attempt**, instead of retrying up to 10 times as intended.

## Checklist

1. Closes #182 

2. How was this tested:
Manually tested by running the workflow and verifying that it retries 10 times before continuing as new. Also validated expected logging behavior.

3. Any docs updates needed?
No updates needed.

### DIFF
--- a/polling/periodic_sequence/workflows.py
+++ b/polling/periodic_sequence/workflows.py
@@ -38,6 +38,7 @@ class ChildWorkflow:
             except ActivityError:
                 workflow.logger.error("Activity failed, retrying in 1 seconds")
             await asyncio.sleep(1)
-            workflow.continue_as_new(name)
 
-        raise Exception("Polling failed after all attempts")
+        workflow.continue_as_new(name)
+        # unreachable
+        return ""


polling/periodic_sequence/workflows.py
```diff
@@ -38,6 +38,7 @@ async def run(self, name: str) -> str:
             except ActivityError:
                 workflow.logger.error("Activity failed, retrying in 1 seconds")
             await asyncio.sleep(1)
-            workflow.continue_as_new(name)
 
-        raise Exception("Polling failed after all attempts")
+        workflow.continue_as_new(name)
+        # unreachable
+        return ""
```
cretz, line 44:
```suggestion
```

`continue_as_new` return type hint has https://docs.python.org/3/library/typing.html#typing.NoReturn, so shouldn't need this part

cretz (commented):
Thanks! Minor suggestion, then happy to merge

polling/periodic_sequence/workflows.py
Nomango: I've made the changes. Let's have the CI workflows check it.

cretz (approved):
Thanks!

PR #186, opened 2025-05-12
Properly skip infrequent polling test on time-skipping server

### LINKED ISSUE #184: [Bug] Infrequent polling test not properly skipped during time skipping
### Describe the bug

#152 added `test_infrequent_polling_workflow`, but put

```python
    if not env.supports_time_skipping:
        pytest.skip("Too slow to test with time-skipping disabled")
```

when likely meant

```python
    if env.supports_time_skipping:
        pytest.skip("Too slow to test with time-skipping disabled")
```

cretz:
## What was changed

Properly skip infrequent polling test on time-skipping server

## Checklist

1. Closes #184

### DIFF
--- a/tests/polling/infrequent/workflow_test.py
+++ b/tests/polling/infrequent/workflow_test.py
@@ -10,7 +10,7 @@ from polling.infrequent.workflows import GreetingWorkflow
 
 
 async def test_infrequent_polling_workflow(client: Client, env: WorkflowEnvironment):
-    if not env.supports_time_skipping:
+    if env.supports_time_skipping:
         pytest.skip("Too slow to test with time-skipping disabled")
 
     # Start a worker that hosts the workflow and activity implementations.


yuandrew (approved):

tests/polling/infrequent/workflow_test.py
```diff
@@ -10,7 +10,7 @@
 
 
 async def test_infrequent_polling_workflow(client: Client, env: WorkflowEnvironment):
-    if not env.supports_time_skipping:
+    if env.supports_time_skipping:
         pytest.skip("Too slow to test with time-skipping disabled")
```
dandavison, line 14: The boolean condition and the message don't seem to jointly make sense any longer?

cretz: Ah, yes, it looks like the original message should have been "enabled" (in addition to the conditional), will fix that

cretz: Hrmm, maybe I misunderstood the situation here and we are wanting to only run this test with time skipping and it's still too slow even with time skipping. Will look.

PR #187, opened 2025-05-23
Change_log_level

GSmithApps:
## What was changed
Added a hello sample showing how to change the log level for workflow task failures

## Why?
Some users have requested a sample on this.

## Checklist

1. Any docs updates needed?
Updated the top-level readme and the hello readme

2. How was this tested:
Ran and confirmed that the log level was changed to error

image


hello/hello_change_log_level.py
```diff
@@ -0,0 +1,61 @@
...
+logging.basicConfig(
+    stream=sys.stdout,
+    level=logging.INFO,
+    format="%(asctime)s  %(levelname)-8s  %(name)s  %(message)s",
+)
+
+
+class CustomLogFilter(logging.Filter):
+    def filter(self, record):
+        msg = record.getMessage()
```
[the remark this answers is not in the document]
GSmithApps: Love learning something new -- thanks for the suggestion!

hello/hello_change_log_level.py
```diff
@@ -0,0 +1,64 @@
...
+import asyncio
+import logging
+import sys
+
+from temporalio import workflow
+from temporalio.client import Client
+from temporalio.worker import Worker
+
+# --- Begin logging set‑up ----------------------------------------------------------
+_WORKFLOW_TASK_FAILURE_LOG_PREFIX = "Failed activation on workflow"
```
GSmithApps, line 14: @cretz I'm a little worried about this hardcoding cuz I'm not sure we guarantee this log message won't be modified. 

- If it won't be modified, I think we don’t have anything to worry about
- if it could be modified, could we do something like expose a constant from the SDK? I recognize it might be a little goofy, but it might be more robust

cretz: Yes, this is exactly why we just solved https://github.com/temporalio/sdk-python/issues/864 last week. I'd wait until that is released (soon).

hello/hello_change_log_level.py
```diff
@@ -0,0 +1,64 @@
...
+    client = await Client.connect("localhost:7233")
+    async with Worker(
+        client,
+        task_queue="hello-task-queue",
+        workflows=[GreetingWorkflow],
+    ):
+        await client.execute_workflow(
+            GreetingWorkflow.run,
+            id="hello-workflow-id",
+            task_queue="hello-task-queue",
```
cretz, line 59: Would recommend an ID and task queue named specifically for this sample

cretz (commented):
LGTM, nothing really blocking, but let's wait for the impending Python SDK release so you can use a more deliberate differentiator on the log record.

hello/hello_change_log_level.py
GSmithApps: Love it - fixed 👍

hello/hello_change_log_level.py
GSmithApps: I see the new SDK was [released](https://github.com/temporalio/sdk-python/releases/tag/1.13.0) and includes the fix (888). Judging by [the files changed in the PR](https://github.com/temporalio/sdk-python/pull/888/files), this string didn’t change, correct? So the benefit of the PR is adding the test to reinforce the fact that this string is public API and won’t change? Meaning, the sample is still good? I know you said nothing blocking

cretz: The message string didn't change, but the purpose of the issue is to add another, more stable string for this use case. So we should encourage in this sample to check `__temporal_error_identifier` attribute instead.

GSmithApps: Thanks — that looked private to me, but upon further inspection, it’s okay to use. Done ✅

GSmithApps:
I believe all comments are resolved and this is ready. The tests are failing, but I suspect it’s due to some flakes because they're failing in `main` too

GSmithApps:
@cretz ready if you are 👍

cretz (approved):
I still think this is a bit too specific of a need to go with all the other hello things, but meh

PR #188, opened 2025-05-29
Added support for LangSmith tracing across workflows and activities

mfateev:
## What was changed
Added an interceptor that propagates LangSmith tracing headers as well as creates corresponding spans for Temporal workflows and activities.

## Why?
Enable LangSmith visibility for workflows that use LangChain.


langchain/langchain_interceptor.py
```diff
@@ -0,0 +1,174 @@
+from __future__ import annotations
+import asyncio
+from contextvars import copy_context
+
+from langsmith import get_tracing_context
```
cretz, line 5: Can this dependency be explicitly added to the `langchain` dependency group in `pyproject.toml`? (even though it is probably transitively included)

cretz, line 5: Importing non-stdlib things as non-passed-through will increase per-workflow memory. Can we move this to inside pass through block?

langchain/langchain_interceptor.py
```diff
@@ -0,0 +1,174 @@
...
+with workflow.unsafe.imports_passed_through():
+    from contextlib import contextmanager
+    from typing import Any, Mapping, Protocol, Type
+
+    import temporalio.activity
+    import temporalio.api.common.v1
+    import temporalio.client
+    import temporalio.converter
+    import temporalio.worker
+    import temporalio.workflow
```
cretz, line 18: Standard library and `temporalio` modules are automatically passed through, they do not need to be under here

langchain/langchain_interceptor.py
```diff
@@ -0,0 +1,174 @@
...
+            if defn is None or defn.name is None:
+                name = "unknown"
+            name = defn.name
+
+        with context_from_header(input, temporalio.workflow.payload_converter()):
+            # This is a sandbox friendly way to write
+            # with trace(...):
+            #   return await self.next.execute_workflow(input)
+            with temporalio.workflow.unsafe.sandbox_unrestricted():
+                t = trace(name=f"execute_workflow:{name}", run_id=workflow.info().run_id)
```
cretz, line 139: Any concerns with tracing during replay here?

langchain/langchain_interceptor.py
cretz, line 1: I see we never really had tests for the `langchain` sample, so don't have to add as part of this, but we should probably make an issue to add them at some point. We've found cases where these integration-type examples become stale or stop working with new dependency versions and there are no tests to confirm that.

langchain/langchain_interceptor.py
mfateev: I didn't see any duplicated traces in the UI when the replay was forced.

langchain/langchain_interceptor.py
mfateev: I'm not sure how a local test would help, as they don't have an embeddable version of Langsmith.

cretz: That is unfortunate. Ideally we'd test the same way they encourage users to test using the in-memory representations or mocks they suggest.

langchain/workflow.py
```diff
@@ -1,17 +1,54 @@
+import asyncio
 from datetime import timedelta
+from typing import List
 
 from temporalio import workflow
 
 with workflow.unsafe.imports_passed_through():
+    from dataclasses import dataclass
```
cretz, line 8: Standard library packages re already passed through

cretz (approved):
LGTM, @dandavison - do you want to add anything?

PR #194, opened 2025-06-11
Add sample workflow for reauthentication after risk event

pankajtaneja5:
## 🛠 What was changed

Added a new sample workflow: `ReauthenticationAfterRiskEventWorkflow`.  
This workflow demonstrates an adaptive authentication pattern where users are required to reauthenticate after suspicious activity — such as login from a new device or after completing account recovery.  

The worker listens for a `reauth_complete` signal, and the workflow can be configured with a timeout to handle cases where reauthentication isn't completed in time.

## 🔐 Why?

This use case is relevant for security-conscious applications that need to enforce reauthentication following risk events.  
The sample highlights how Temporal can model secure, stateful authentication workflows that are durable, auditable, and cleanly separated from business logic.

It aligns with modern security practices around risk-based session validation and account protection.

## ✅ Checklist

- [x] New sample added to `samples-python`
- [x] Demonstrates secure workflow pattern using Temporal signals and timeouts
- [x] Includes worker and starter code for demonstration

## Notes

There wasn’t a separate issue filed for this PR.  
If preferred, I can open one and link it here for tracking purposes.

Let me know if any adjustments are needed!

### DIFF
--- a/reauth_after_risk_event/starter.py
+++ b/reauth_after_risk_event/starter.py
@@ -1,0 +1,19 @@
+import asyncio
+from datetime import timedelta
+from temporalio.client import Client
+
+async def main():
+    client = await Client.connect("localhost:7233")
+
+    result = await client.start_workflow(
+        "ReauthenticationAfterRiskEventWorkflow",
+        "user-123",  # user_id
+        id="reauth-workflow-user-123",
+        task_queue="reauth-task-queue",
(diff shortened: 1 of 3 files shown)


brianstrauch:
Closing this out — it's been in draft for over a year with no CLA signature.

The pattern here is essentially a signal with a timeout, which is already covered by `hello/hello_signal` and the samples under `message_passing/`.

PR #200, opened 2025-06-22
port mutex sample from go repository

K-Yo:
## What was changed

Added a python sample for mutex pattern

## Why?

The sample exists for [go sdk](https://github.com/temporalio/samples-go/blob/main/mutex) but not for python yet. I needed it so I figured I could share it.

## Checklist

1. TODO

I’ve left a TODO in the code, I am not sure how to properly distinguish who sends a release in python (even if sender workflows are supposed to send them in order)

3. How was this tested:

ran it locally with latest temporal dev server, and using branch pyproject.toml configuration.

4. Any docs updates needed?

not that I think of

### DIFF
--- a/mutex/mutexworkflow.py
+++ b/mutex/mutexworkflow.py
@@ -1,0 +1,117 @@
+from dataclasses import dataclass
+from temporalio.client import Client
+from temporalio import activity, workflow
+import asyncio
+from temporalio.exceptions import ApplicationError
+
+LOCK_ACQUIRED_SIGNAL_NAME = "acquire-lock-event"
+
+
+@dataclass
+class SignalWithStartMutexWorkflowInput:
+    namespace: str
+    resource_id: str
+    unlock_timeout_seconds: float
+    sender_workflow_id: str
+
+
+@dataclass
+class SignalWithStartMutexWorkflowResult:
+    workflow_id: str
+
+
+@dataclass
+class MutexWorkflowInput:
+    namespace: str
+    resource_id: str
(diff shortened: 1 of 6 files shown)


mutex/mutexworkflow.py
```diff
@@ -0,0 +1,117 @@
...
+        start_signal_args=[params.sender_workflow_id],
+    )
+    return SignalWithStartMutexWorkflowResult(workflow_id=workflow_id)
+
+
+def generate_unlock_token(sender_workflow_id: str) -> str:
+    return f"unlock-event-{sender_workflow_id}"
+
+
+@workflow.defn
```
cretz, line 60: I think we may not want to blind-port this from Go but have a more Pythonic way to do a mutex and we want to make sure workflows are in their own files separate from activities/models. For instance, the .NET version of this sample at https://github.com/temporalio/samples-dotnet/tree/main/src/Mutex made an idiomatic `WorkflowMutex` class for general purpose use. I think this sample should consider a workflow-backed `Lock` class that looks almost exactly like https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock and is copyable/usable as if it is a library.

p44ytwz25w-cloud: write permissions required into the code before lock can be engaged

brianstrauch:
@K-Yo — checking in here, sorry for the long delay. Still interested in reworking it? If we don't hear back in the next couple of weeks we'll close it out, and you're welcome to reopen whenever.

Issue #190, opened 2025-06-03
[Feature Request] Remove default groups from pyproject.toml


cretz:
### Describe the solution you'd like

Default groups were inadvertently added when we switched from Poetry to uv. Remove them.

cretz:
To clarify, trying to solve the case where non-shared/unrelated dependency groups are installed by default now, whereas they weren't before

PR #201, opened 2025-06-23
Expense example

jssmith:
## What was changed

Ported the expense example from Go to Python

## Why?

Demonstrate human-in-the-loop processing and asynchronous activity completion

## Checklist

1. Closes N/A

2. How was this tested:
Includes tests

3. Any docs updates needed?
README is updated


expense/activities.py
```diff
@@ -0,0 +1,89 @@
...
+
+from expense import EXPENSE_SERVER_HOST_PORT
+
+
+@activity.defn
+async def create_expense_activity(expense_id: str) -> None:
+    if not expense_id:
+        raise ValueError("expense id is empty")
+
+    async with httpx.AsyncClient() as client:
```
cretz, line 12: Consider making the activities as methods on a class that instantiates this in `__init__` and reuses it. It is a bit better performance to reuse an HTTP client/connection and I think a better practice to demonstrate to users

expense/activities.py
```diff
@@ -0,0 +1,89 @@
...
+        response = await client.get(
+            f"{EXPENSE_SERVER_HOST_PORT}/create",
+            params={"is_api_call": "true", "id": expense_id},
+        )
+        response.raise_for_status()
```
cretz, line 17: _Technically_ 4xx is probably a `ApplicationError` with `non_retryable=True`, but that's a bit pedantic. But with this setup, an activity that, say, has invalid auth will retry forever

expense/activities.py
```diff
@@ -0,0 +1,89 @@
...
+    if status == "SUCCEED":
+        # register callback succeed
+        logger.info(f"Successfully registered callback. ExpenseID: {expense_id}")
+
+        # Raise the complete-async error which will return from this function but
+        # does not mark the activity as complete from the workflow perspective.
+        #
+        # Activity completion is signaled in the `notify_expense_state_change`
+        # function in `ui.py`.
+        activity.raise_complete_async()
```
cretz, line 66: I see 8 years ago in the Go samples we used async activity completion. But another alternative that may be more popular these days is sending a signal on completion. See https://docs.temporal.io/activity-execution#when-to-use-async-completion. In this case, unless that external call is heartbeating and can react to cancellation, signal may be best. But not a big deal.

expense/workflow.py
```diff
@@ -0,0 +1,56 @@
...
+
+        # Step 1: Create new expense report
+        try:
+            await workflow.execute_activity(
+                create_expense_activity,
+                expense_id,
+                start_to_close_timeout=timedelta(seconds=10),
+            )
+        except Exception as err:
+            logger.error(f"Failed to create expense report: {err}")
```
cretz, line 27: Change to `logger.exception` if you want the exception info like message and stack trace automatically added

pyproject.toml
```diff
@@ -64,13 +64,19 @@ cloud-export-to-parquet = [
...
+    "uvicorn[standard]>=0.24.0.post1,<0.25",
+]
 
 [tool.uv]
 default-groups = [
     "dev",
     "bedrock",
     "dsl",
     "encryption",
+    "expense",
```
cretz, line 79: We accidentally added these default groups and we will remove them in #190

cretz (approved):
Only minor things, nothing technically blocking

expense/activities.py
jssmith: I think signal makes more sense—particularly since heartbeating doesn't really make sense when sending for human review. I like keeping the samples consistent, but really we should go back and fix the Go.

PR #211, opened 2025-07-17
Minor cleanups: interceptors and OpenAI Agents SDK

dandavison:


openai_agents/workflows/research_agents/research_manager.py
```diff
@@ -46,16 +46,13 @@ async def _plan_searches(self, query: str) -> WebSearchPlan:
 
     async def _perform_searches(self, search_plan: WebSearchPlan) -> list[str]:
         with custom_span("Search the web"):
-            num_completed = 0
```
tconley1428, line 49: Do we want to do code cleanup which makes it different from the openai sample?

dandavison: No strong opinion personally. I didn't know this code was copied.

dandavison: Reverted. Doesn't seem much point if it's copied code.

tconley1428 (approved):

PR #226, opened 2025-07-27
OpenAI Agents financial research and reasoning examples

jssmith:
## What was changed

Ported financial research and reasoning examples from OpenAI Agents SDK.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/openai_agents/handoffs/README.md
+++ b/openai_agents/handoffs/README.md
@@ -1,0 +1,44 @@
+# Handoffs Examples
+
+Agent handoff patterns with message filtering in Temporal workflows.
+
+*Adapted from [OpenAI Agents SDK handoffs examples](https://github.com/openai/openai-agents-python/tree/main/examples/handoffs)*
+
+Before running these examples, be sure to review the [prerequisites and background on the integration](../README.md).
+
+## Running the Examples
+
+First, start the worker:
+```bash
+uv run openai_agents/handoffs/run_worker.py
+```
+
+Then run the workflow:
+
+### Message Filter Workflow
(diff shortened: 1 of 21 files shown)


cretz (commented):
Currently, the top-level of `openai_agents` has all of the samples mashed together as one project. I think it makes the addition of this as a subdir a bit confusing. I wonder if we should move the current sample set into a subdir (e.g. "simple" or something) or make each sample its own dir as we are doing with these newer ones.

openai_agents/financial_research_agent/README.md
cretz, line 1: Currently, the top-level of `openai_agents` has all of the samples mashed together as one project. I think it makes the addition of this as a subdir a bit confusing. I wonder if we should move the current sample set into a subdir (e.g. "simple" or something) or make each sample its own dir as we are doing with these newer ones.

cretz, line 1: I am a little concerned we are adding _lots_ of OpenAI sample code with no tests to confirm it continues to work. Granted I know this is unfortunately OpenAI's approach to samples, but ideally we can have some test somewhere. We have other samples where we have done this and it comes back to bite us hard when people report it stops working and we have to scramble (this is the case with the Sentry sample which has now remained broken for a long time).

tconley1428 (approved):

openai_agents/financial_research_agent/README.md
cretz: :+1: Should we open an issue to track this?

Issue #178, opened 2025-04-21
[Feature Request] External client configuration


cretz:
### Describe the solution you'd like

Update all samples to leverage external client configuration once https://github.com/temporalio/sdk-python/issues/835 is complete.

PR #228, opened 2025-07-29
External client configuration sample

THardy98:
## What was changed
Basic samples for environment configuration, showing default usage and usage with a specific profile and environment variables overrides.

Partly addresses #178

2. How was this tested:
Samples run correctly

3. Any docs updates needed?
I don't think so


THardy98:
Would like to wait for https://github.com/temporalio/sdk-core/pull/968 before changing the existing samples, so we can simplify the diff (otherwise we'll need to read & load from a config file)

env_config/README.md
cretz, line 1: I assume the updating of all existing samples to leverage the one-liner for loading from env config will be a separate PR?

env_config/load_profile.py
```diff
@@ -0,0 +1,56 @@
...
+
+    config_file = Path(__file__).parent / "config.toml"
+    profile_name = "staging"
+
+    # In a real application, these would be set in your shell or deployment
+    # environment (e.g., `export TEMPORAL_ADDRESS=localhost:7233`).
+    # For this sample, we pass them as a dictionary to demonstrate.
+    override_env = {
+        "TEMPORAL_ADDRESS": "localhost:7233",
+    }
```
cretz, line 27: I don't think this is how we should encourage people to do overrides. Rather we should encourage them to alter the returned `connect_config`. They shouldn't mock environment variables.

env_config/README.md
cretz, line 1: Don't forget to update main README with a link to this sample

cretz (approved):

PR #237, opened 2025-08-22
fix worker task queue name in basic example

jinjiaKarl:
## What was changed

fix worker task queue name in basic example, otherwise the workflow cannot be running

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


tconley1428:
Good catch, thank you.

PR #245, opened 2025-09-13
OpenAI Agents Sessions on PostgreSQL

jssmith:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/openai_agents/memory/postgres_session.py
+++ b/openai_agents/memory/postgres_session.py
@@ -1,0 +1,286 @@
+from __future__ import annotations
+
+import json
+from typing import Any
+
+import asyncpg
+from temporalio import activity, workflow
+from pydantic import BaseModel
+from agents.memory.session import SessionABC
+from agents.items import TResponseInputItem
+from openai_agents.memory.db_utils import IdempotenceHelper
+from typing import Callable
+
+_connection_factory: Callable[[], asyncpg.Connection] | None = None
+
+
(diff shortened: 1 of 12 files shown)


openai_agents/memory/postgres_session.py
cretz, line 1: Does OpenAI provide any session implementations themselves here? Can we just write a `DurableSession` that wraps those instead of having to do all of the DB stuff ourselves? A `DurableSession` that accepts a user-provided session means they won't be limited to just specific implementations we implement.

openai_agents/memory/postgres_session.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def __init__(
+        self,
+        session_id: str,
+        config: PostgresSessionConfig,
+    ):
+        self.session_id = session_id
+        self.config = config
+
+    async def get_items(self, limit: int | None = None) -> list[TResponseInputItem]:
```
cretz, line 199: If all of these items will end up in history anyways as a result of this call, is there any value in using a separate database instead of in-memory?

openai_agents/memory/postgres_session.py
cretz: Or [SQLiteSession](https://github.com/openai/openai-agents-python/blob/e486b3a6bdc90fe1826944b68c22ac3b559fb6dc/src/agents/memory/sqlite_session.py) since this is just a sample

PR #252, opened 2025-09-28
nexus_sync_operations

dandavison:
See https://github.com/temporalio/features/issues/664


nexus_sync_operations/caller/workflows.py
```diff
@@ -0,0 +1,41 @@
...
+        nexus_client = workflow.create_nexus_client(
+            service=GreetingService,
+            endpoint=NEXUS_ENDPOINT,
+        )
+
+        # Get supported languages
+        supported_languages = await nexus_client.execute_operation(
+            GreetingService.get_languages, GetLanguagesInput(include_unsupported=False)
+        )
+        print(f"supported languages: {supported_languages}")
```
cretz, line 30: Would discourage use of `print` in workflows

nexus_sync_operations/endpoint_description.md
cretz, line 1: I was under the impression we are only doing this "description" markdown on the "hello" Nexus sample? Do we need to start doing this for all samples?

nexus_sync_operations/handler/service_handler.py
```diff
@@ -0,0 +1,81 @@
...
+    # query against a long-running workflow that is private to the nexus service.
+    @nexusrpc.handler.sync_operation
+    async def get_language(
+        self, ctx: nexusrpc.handler.StartOperationContext, input: None
+    ) -> Language:
+        return await self.greeting_workflow_handle.query(GreetingWorkflow.get_language)
+
+    # 👉 This is a handler for a nexus operation whose internal implementation involves executing an
+    # update against a long-running workflow that is private to the nexus service.
+    @nexusrpc.handler.sync_operation
```
cretz, line 73: In theory an update can be long running and we should demonstrate async operation with it, but I understand if we need some kind of server side support for update callbacks or something for that. May be worth noting though, up to you.

PR #268, opened 2025-12-17
Add Workflow URL to Temporal Sentry errors

Natim:
## What was changed
Add Workflow URL to Temporal Sentry errors

## Why?
It allows to get the Temporal UI workflow URL in the context of the Sentry

image

### DIFF
--- a/sentry/interceptor.py
+++ b/sentry/interceptor.py
@@ -15,8 +15,16 @@ from temporalio.worker import (
 with workflow.unsafe.imports_passed_through():
     import sentry_sdk
 
-
 logger = logging.getLogger(__name__)
+
+TEMPORAL_UI_URL = "https://cloud.temporal.io"
+
+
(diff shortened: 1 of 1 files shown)


sentry/interceptor.py
```diff
@@ -35,6 +43,27 @@ async def execute_activity(self, input: ExecuteActivityInput) -> Any:
                 "temporal.workflow.namespace", activity_info.workflow_namespace
             )
             scope.set_tag("temporal.workflow.run_id", activity_info.workflow_run_id)
+
+            # Build workflow URL for easy navigation from Sentry
```
cretz: This is meant to be a sample a user can adjust with these more advanced needs rather than a library. I am not sure we want the sample assuming anything about cloud, though it would totally make sense for some users to set this tag (or any other tag) when they copy this sample and adjust it for their needs.

PR #286, opened 2026-04-03
Add lambda worker sample

Sushisource:
This PR adds a sample for the new lambda worker usable in conjunction with the new serverless workers features on server/cloud.

It also introduces this sample as a completely independent Python package. All other samples will be converted to work this way in the near future.


lambda_worker/workflows.py
```diff
@@ -0,0 +1,22 @@
...
+
+from temporalio import common, workflow
+
+TASK_QUEUE = "serverless-task-queue-python"
+
+with workflow.unsafe.imports_passed_through():
+    from activities import hello_activity
+
+
+@workflow.defn(versioning_behavior=common.VersioningBehavior.AUTO_UPGRADE)
```
lennessyy: I seem to remember hearing at the demo meeting that we recommend using `PINNED` instead of auto-upgrade. Should we do the same in the samples?

Sushisource: Yeah good point, thanks for the reminder.

tconley1428 (approved):

THardy98 (approved):

PR #288, opened 2026-04-10
AI-18: Move set_tracing_disabled from workflow to worker startup

### LINKED ISSUE #243: [Bug] Remove/move set_tracing_disabled from inside OpenAI sample workflows
### Describe the bug

Two workflows at https://github.com/temporalio/samples-python/tree/main/openai_agents/model_providers/workflows encourage users to call the global `set_tracing_disabled` _inside_ the workflow. Should either remove altogether or move to the `_worker.py` file instead.

xumaple:
## Summary

- Moved `set_tracing_disabled(disabled=True)` out of workflow `run()` methods and into worker `main()` functions in the `openai_agents/model_providers/` samples
- `set_tracing_disabled` is a global side effect that shouldn't live inside deterministic workflow code — it belongs at worker startup where it runs once per process

Fixes #243

### DIFF
--- a/openai_agents/model_providers/run_gpt_oss_worker.py
+++ b/openai_agents/model_providers/run_gpt_oss_worker.py
@@ -27,6 +27,9 @@ class CustomModelProvider(ModelProvider):
 
 
 async def main():
+    # Disable Agents SDK tracing — the default exporter sends traces to OpenAI's
+    # backend, which requires an OpenAI API key not available in these samples.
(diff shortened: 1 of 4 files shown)


DABH (approved):
Yep, this does exactly what #243 suggests.  Thanks for the fix and for the comments that explain why `set_tracing_disabled` is now being called in the right place.

tconley1428 (approved):

PR #290, opened 2026-04-15
Sample code for Nexus messaging

Evanthx:
This is sample code to show two ways to send messages (signals, queries, and updates) through Nexus.


nexus_messaging/callerpattern/caller/workflows.py
```diff
@@ -0,0 +1,74 @@
+"""
+A caller workflow that executes Nexus operations. The caller does not have information
+about how these operations are implemented by the Nexus service.
+"""
+
+from temporalio import workflow
+from temporalio.exceptions import ApplicationError
+
+with workflow.unsafe.imports_passed_through():
```
VegetarianOrc, line 9: No need to use `imports_passed_through()` here.

nexus_sync_operations/caller/workflows.py
VegetarianOrc, line 1: Was this same intentionally deleted? I don't think we should remove the basic sync operation example in favor of a messaging sample.

nexus_messaging/callerpattern/handler/workflows.py
```diff
@@ -0,0 +1,88 @@
...
+details of the Nexus service: the caller only interacts via Nexus operations.
+"""
+
+import asyncio
+from datetime import timedelta
+
+from temporalio import workflow
+from temporalio.exceptions import ApplicationError
+
+with workflow.unsafe.imports_passed_through():
```
VegetarianOrc, line 13: Same as above. Shouldn't need `imports_passed_through()` here.

nexus_messaging/callerpattern/README.md
```diff
@@ -0,0 +1,56 @@
+## Caller pattern
+
+The handler worker starts a `GreetingWorkflow` for a user ID.
+`NexusGreetingServiceHandler` holds that ID and routes every Nexus operation to it.
+The caller's input does not have that workflow ID as the caller doesn't know it -- but the caller
+sends in the User ID, and `NexusGreetingServiceHandler` knows how to get the desired workflow ID
+from that User ID (see the `get_workflow_id` call).
+
+The handler worker uses the same `get_workflow_id` call to generate a workflow ID from a user ID
+when it launches the workflow.
```
VegetarianOrc, line 10: I'm worried this sample is a little complicated.

For example, if our goal is to demonstrate how to acquire a workflow handle in a nexus operation handler and signal/query/update a workflow I think abstracting workflow ID as a combo of prefix + userID just adds cognitive load and may be confusing.

If the goal is to show how Nexus can be used to abstract the underlying workflow away, then I think that might deserve a sample separate from how to use messaging via Nexus.

nexus_messaging/callerpattern/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+The caller workflow:
+1. Queries for supported languages (`get_languages` -- backed by a `@workflow.query`)
+2. Changes the language to Arabic (`set_language` -- backed by a `@workflow.update` that calls an activity)
+3. Confirms the change via a second query (`get_language`)
+4. Approves the workflow (`approve` -- backed by a `@workflow.signal`)
```
VegetarianOrc, line 16: Nit: The caller workflow only executes nexus operations and it feels a bit confusing to call out that they're backed by workflow.query/update/signal. IMO that info should be called out about or alongside the service handler.

nexus_messaging/callerpattern/handler/worker.py
```diff
@@ -0,0 +1,62 @@
...
+    # Start the long-running entity workflow that backs the Nexus service,
+    # if not already running.
+    workflow_id = get_workflow_id(USER_ID)
+    await client.start_workflow(
+        GreetingWorkflow.run,
+        id=workflow_id,
+        task_queue=TASK_QUEUE,
+        id_conflict_policy=WorkflowIDConflictPolicy.USE_EXISTING,
+    )
+    logging.info("Started greeting workflow: %s", workflow_id)
```
VegetarianOrc, line 42: Nit: this just starts a workflow while no workers are running. I'd consider moving it in the below `async with Worker(...)` block

nexus_messaging/callerpattern/handler/service_handler.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Routes to set_language_using_activity (not set_language) so that new languages not
+    # already in the greetings map can be fetched via an activity.
+    @nexusrpc.handler.sync_operation
+    async def set_language(
+        self, ctx: nexusrpc.handler.StartOperationContext, input: SetLanguageInput
+    ) -> Language:
+        return await self._get_workflow_handle(input.user_id).execute_update(
+            GreetingWorkflow.set_language_using_activity, input
+        )
```
VegetarianOrc, line 71: If `GreetingWorkflow.set_language` is unused, we should remove it IMO.

nexus_messaging/ondemandpattern/README.md
```diff
@@ -0,0 +1,60 @@
+## On-demand pattern
+
+No workflow is pre-started. The caller creates and controls workflow instances through Nexus
+operations. `NexusRemoteGreetingService` adds a `run_from_remote` operation that starts a new
+`GreetingWorkflow`, and every other operation includes a `user_id` so the handler knows which
+instance to target.
```
VegetarianOrc, line 6: Can you help me understand the motivation behind the two samples? This sample appears to demonstrate the same thing as the above sample in regards to messaging via Nexus.

The messaging via Nexus part remains using a sync operation to:
- determine workflow ID
- execute signal/query/update

The Nexus operations are not opinionated about where the workflow started from.

VegetarianOrc (commented):
Code looks okay to me! A few comments throughout. 

Left a couple of comments to help me understand the motivation for the structure of the samples as well.

Would like to clear up the intention around removing the sync operation sample before approving.

nexus_messaging/callerpattern/handler/workflows.py
Evanthx: Done

nexus_messaging/callerpattern/caller/workflows.py
Evanthx: Done

nexus_sync_operations/caller/workflows.py
Evanthx: So this was the original base that I used as a starting point as it had messaging - everything that was in it is now in the new sample. I just moved the package name to be consistent with what I was doing elsewhere.

You have a point about basic sync operations - but this already had messaging in it so it was already not a basic sync example? So I think this is OK, but I can also see it being debatable, so leaving this comment unresolved so you can reply.

nexus_messaging/callerpattern/handler/worker.py
Evanthx: Done

nexus_messaging/callerpattern/README.md
Evanthx: This was actually done as a result of a code review in the Java sample as it was thought that exposing the workflow ID was not desirable!

nexus_messaging/callerpattern/README.md
Evanthx: Since this is demonstrating queries, updates and signals this does actually feel useful to me though

nexus_messaging/callerpattern/handler/service_handler.py
Evanthx: It is being used but calls an activity. --honestly this is what was in the original code that I extended, and was mostly just left alone!

nexus_messaging/ondemandpattern/README.md
Evanthx: This happened after several conversations with Quinn. How the caller works is slightly different based on whether it launches the workflow or not, so we ended up doing both ways. A lot of the differences are in how the workflowID is handled, which you actually called out in an earlier message - but that is what we were trying to demonstrate.

nexus_messaging/callerpattern/handler/worker.py
Evanthx: actually didn't do - the worker is a sync call, so it won't go past the worker call until it's done. So I could make it an async call, but seemed simpler the way it was!

nexus_sync_operations/caller/workflows.py
VegetarianOrc: As long as we update the [related doc](https://docs.temporal.io/develop/python/nexus/feature-guide#use-the-temporal-client-for-signals-queries-and-updates) about this to reference this new sample, I think it's probably fine given the context!

In fact, we probably should take a pass on those docs for each lang as these samples get merged in to ensure they reflect the new sample, links are valid, etc.

nexus_messaging/ondemandpattern/handler/workflows.py
```diff
@@ -0,0 +1,89 @@
...
+operations.
+"""
+
+import asyncio
+from datetime import timedelta
+
+from temporalio import workflow
+from temporalio.exceptions import ApplicationError
+
+with workflow.unsafe.imports_passed_through():
```
VegetarianOrc: Looks like I missed at least one more of these in the first pass. Please take a pass and make sure we get 'em all 😄

VegetarianOrc (approved):

nexus_messaging/ondemandpattern/handler/workflows.py
Evanthx: I didn't think to search for them - sorry about that!

Evanthx: Done

nexus_sync_operations/caller/workflows.py
Evanthx: Oh good catch, I didn't even think about that!!

Evanthx: Docs are updated, will submit that once the sample code is in.

PR #292, opened 2026-04-17
Add LangSmith tracing sample

xumaple:
## Summary
- Adds `langsmith_tracing/` sample demonstrating `temporalio.contrib.langsmith.LangSmithPlugin`
- **basic/**: one-shot LLM workflow (prompt → OpenAI activity → result)
- **chatbot/**: conversational loop with `save_note`/`read_note` tools, signals, queries, and dynamic LangSmith trace names
- Includes mocked tests, README, and `langsmith-tracing` dependency group
- Declares `langchain` and `langsmith-tracing` as conflicting groups in `[tool.uv]` to avoid version conflicts

## Test plan
- [ ] `poe format` — passes
- [ ] `mypy` on `langsmith_tracing/` and `tests/langsmith_tracing/` — passes
- [ ] `pytest tests/langsmith_tracing/` — basic test passes, chatbot tests skip (pending SDK plugin release)
- [ ] Manual test: run worker + starter for both basic and chatbot examples


langsmith_tracing/basic/activities.py
```diff
@@ -0,0 +1,31 @@
...
+    instructions: str = "You are a helpful assistant."
+
+
+@traceable(name="Call OpenAI", run_type="llm")
+@activity.defn
+async def call_openai(request: OpenAIRequest) -> Response:
+    """Call OpenAI Responses API. Retries handled by Temporal, not the OpenAI client."""
+    # wrap_openai patches the client so each API call (e.g. responses.create)
+    # creates its own child span with model parameters and token usage.
+    client = wrap_openai(AsyncOpenAI(max_retries=0))
```
JasonSteving99, line 25: maybe worth a comment explaining why retries set to 0 (since temporal activity handles the retries)

langsmith_tracing/basic/README.md
```diff
@@ -0,0 +1,50 @@
...
+See the [parent README](../README.md) for prerequisites.
+
+## Running
+
+```bash
+# Terminal 1 — start the worker
+uv run --group langsmith-tracing python -m langsmith_tracing.basic.worker
+
+# Terminal 2 — run the workflow
+uv run --group langsmith-tracing python -m langsmith_tracing.basic.starter
```
brianstrauch, line 14: I've been creating examples where the worker and workflow are run in the same file. It makes the setup instructions more minimal which I think is great for these basic samples. Example: https://github.com/temporalio/sdk-python/pull/1448/changes#diff-0b5b90d5d69d62c65cd6eccd7529d2461386d176c3bace0995172581ce1e31dfR45-R63

langsmith_tracing/basic/worker.py
```diff
@@ -0,0 +1,58 @@
...
+        print("Shutting down")
+
+
+if __name__ == "__main__":
+    loop = asyncio.new_event_loop()
+    try:
+        loop.run_until_complete(main())
+    except KeyboardInterrupt:
+        interrupt_event.set()
+        loop.run_until_complete(loop.shutdown_asyncgens())
```
brianstrauch, line 58: Why all this instead of just the default `asyncio.run(main())` that other examples use?

brianstrauch: I wonder if my earlier comment about combining the worker and workflow in the same file makes handling these interrupts more graceful.

langsmith_tracing/basic/worker.py
```diff
@@ -0,0 +1,58 @@
...
+    # add_temporal_runs=True creates LangSmith runs for each Temporal
+    # workflow/activity execution. False (default) only propagates context.
+    add_temporal_runs = "--add-temporal-runs" in sys.argv
+
+    config = ClientConfig.load_client_connect_config()
+    config.setdefault("target_host", "localhost:7233")
+
+    client = await Client.connect(
+        **config,
+        data_converter=pydantic_data_converter,
```
JasonSteving99, line 31: I wonder if it'd be simpler not to use pydantic for an example like this that doesn't strictly need pydantic? I don't care strongly, I'm open to keeping it if you think it's adding enough value

langsmith_tracing/basic/workflows.py
```diff
@@ -0,0 +1,34 @@
...
+            metadata={"workflow_id": workflow.info().workflow_id},
+            tags=["basic-llm"],
+        )
+        async def _run() -> str:
+            response = await workflow.execute_activity(
+                call_openai,
+                OpenAIRequest(model="gpt-4o-mini", input=prompt),
+                start_to_close_timeout=timedelta(seconds=60),
+            )
+            return response.output_text
```
JasonSteving99, line 32: for prior comment re: pydantic, you could probably skip the pydantic data converter by having the activity just return this `Response::output_text` instead of the whole `Response` since this is all you're actually using

langsmith_tracing/chatbot/starter.py
```diff
@@ -0,0 +1,93 @@
...
+            )
+            async def send_and_wait(msg: str):
+                prev_response = await wf_handle.query(ChatbotWorkflow.last_response)
+                await wf_handle.signal(ChatbotWorkflow.user_message, msg)
+                for _ in range(60):
+                    await asyncio.sleep(0.5)
+                    response = await wf_handle.query(ChatbotWorkflow.last_response)
+                    if response != prev_response:
+                        return response
+                return None
```
JasonSteving99, line 78: maybe worth encapsulating this logic in the workflow itself to demonstrate that a workflow update func can synchronously wait for the workflow to get to a desired state rather than looping over queries here

langsmith_tracing/chatbot/workflows.py
```diff
@@ -0,0 +1,165 @@
...
+
+from temporalio import workflow
+from temporalio.common import RetryPolicy
+
+with workflow.unsafe.imports_passed_through():
+    from langsmith import traceable
+
+    from langsmith_tracing.chatbot.activities import OpenAIRequest, call_openai
+
+RETRY = RetryPolicy(initial_interval=timedelta(seconds=2), maximum_attempts=3)
```
JasonSteving99, line 14: nit: this is only used once, may as well be inlined

langsmith_tracing/chatbot/workflows.py
```diff
@@ -0,0 +1,165 @@
...
+    # wrap an inner function.
+    @workflow.run
+    async def run(self) -> str:
+        now = workflow.now().strftime("%b %d %H:%M")
+        return await traceable(
+            name=f"Session {now}",
+            run_type="chain",
+            metadata={"workflow_id": workflow.info().workflow_id},
+            tags=["chatbot-session"],
+        )(self._session)()
```
JasonSteving99, line 87: Maybe worth a short comment that this is an alternative way to mark funcs traceable rather than the nested func def with the `@traceable(...)` decorator that you mostly used elsewhere

langsmith_tracing/basic/README.md
xumaple: This goes directly against temporal's [official guidance](https://learn.temporal.io/tutorials/python/background-check/project-setup/) on file organization though. It's fine for test code but I don't think we should be doing that for samples.

langsmith_tracing/chatbot/workflows.py
xumaple: True. we can define an inline function for dynamic trace names, but this traceable function is another way.

langsmith_tracing/chatbot/starter.py
xumaple: In terms of what we see in Langsmith, if we put this in workflow update, then it no longer shows up as a child of RunWorkflow, but rather each workflow update would be a separate sibling to Start/RunWorkflow in the form of StartWorkflowUpdate. Is that what you're imagining?

langsmith_tracing/basic/worker.py
xumaple: changed to asyncio.run(main())

langsmith_tracing/chatbot/starter.py
xumaple: obsolete after the change to use updates

langsmith_tracing/chatbot/workflows.py
xumaple: done

langsmith_tracing/chatbot/workflows.py
xumaple: done

langsmith_tracing/basic/workflows.py
xumaple: removed

langsmith_tracing/basic/worker.py
xumaple: removed

langsmith_tracing/basic/activities.py
xumaple: done

brianstrauch (approved):

langsmith_tracing/basic/starter.py
```diff
@@ -0,0 +1,54 @@
...
+        project_name=PROJECT_NAME,
+        tags=["client-side"],
+    )
+    async def run_workflow(prompt: str) -> str:
+        return await client.execute_workflow(
+            BasicLLMWorkflow.run,
+            prompt,
+            id="langsmith-basic-workflow-id",
+            task_queue="langsmith-basic-task-queue",
+        )
```
JasonSteving99, line 47: probably worth just making this func top-level since it doesn't require the `@traceable(...)` decorator closing over anything

JasonSteving99 (approved):
Had one minor suggestion leftover, but LGTM!

PR #297, opened 2026-04-24
Nexus samples: use business IDs for workflow IDs

VegetarianOrc:
## What was changed

Stop using the Nexus request ID (nexus_cancel) or bare uuid4 (hello_nexus, nexus_multiple_args) as the backing workflow ID. Build a meaningful business ID from the operation input.

## Why?

Our samples should reflect best practices around creating meaningful IDs. Part of https://github.com/temporalio/features/issues/692.

### DIFF
--- a/hello_nexus/handler/service_handler.py
+++ b/hello_nexus/handler/service_handler.py
@@ -33,7 +33,7 @@ class MyNexusServiceHandler:
         return await ctx.start_workflow(
             WorkflowStartedByNexusOperation.run,
             input,
-            id=str(uuid.uuid4()),
+            id=f"hello-nexus-workflow-{input.name}-{uuid.uuid4()}",
         )
 
     # This is a Nexus operation that responds synchronously to all requests. That means
--- a/hello_nexus/handler/service_handler.py
+++ b/hello_nexus/handler/service_handler.py
@@ -4,8 +4,6 @@ This file demonstrates how to implement a Nexus service.
 
 from __future__ import annotations
 
-import uuid
-
 import nexusrpc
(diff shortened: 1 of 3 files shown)


Evanthx (approved):
LGTM

chrsmith (approved):
As far as I know this is all on the up and up. But do we still want to include a UUID in the "Workflow ID"? Since it isn't required for uniqueness (that's the "Run ID", right?) wouldn't this just lead to Workflow ID spam?

e.g. you'd want the Workflow ID to be something like "reticulate-splines_job-xxx" (so the workflow ID would be unique per-job, but not necessarily need its own UUID)?

I assume this is the "right way", mostly asking for my own understanding.

VegetarianOrc:
> As far as I know this is all on the up and up. But do we still want to include a UUID in the "Workflow ID"? Since it isn't required for uniqueness (that's the "Run ID", right?) wouldn't this just lead to Workflow ID spam?
> 
> e.g. you'd want the Workflow ID to be something like "reticulate-splines_job-xxx" (so the workflow ID would be unique per-job, but not necessarily need its own UUID)?
> 
> I assume this is the "right way", mostly asking for my own understanding.

The workflow ID is the main tool used to identify a workflow after it's been started. The Run ID is useful for differentiating workflow runs that happen in various scenarios like a workflow that was continued-as-new. You're right that the workflow ID doesn't necessarily need its own UUID and could use some sequential job number or the like to use unique jobs. In many scenarios the differentiator in a workflow ID winds up being a UUID because the underlying business concept is also identified by a UUID (i.e. user id, order id, etc).

PR #299, opened 2026-04-30
Workflow Streams samples 

jssmith:
Add samples to accompany Workflow Streams


jssmith:
Superseded by #300 (workflow_streams basics) and #301 (openai_agents streaming sample), which split this PR into two reviewable halves off a common parent.

PR #300, opened 2026-04-30
Add workflow_streams samples

jssmith:
## Summary

Adds samples for `temporalio.contrib.workflow_streams`, the workflow-hosted durable event stream contrib (still experimental, ships in `temporalio>=1.27.0`).

Four scenarios under `workflow_streams/`, each exercising more of the API:

- `order_workflow` — minimal publisher driven by an activity, with two typed topics on a single stream
- `pipeline_workflow` — reconnecting subscriber across activity steps
- `hub_workflow` — external publisher driving a long-lived workflow
- `ticker_workflow` — truncating publisher with a bounded ring buffer

Workflows bind typed topic handles in `@workflow.init` and publish via `topic.publish(value)`. Subscribers iterate from a `WorkflowStreamClient`. A small `race_with_workflow` helper races the consumer against `handle.result()` so a workflow failure surfaces as an exception rather than blocking the subscriber forever.

This is one half of #299, split out so the workflow_streams basics can land independently of the openai_agents streaming sample (separate PR).

## Test plan

- [ ] `uv run workflow_streams/run_worker.py` and run each `run_*.py` starter; verify expected output
- [ ] Reconnecting subscriber: kill and restart mid-run, confirm resume from prior offset
- [ ] Truncating ticker: confirm ring-buffer drops oldest items as expected


workflow_streams/workflows/hub_workflow.py
```diff
@@ -0,0 +1,40 @@
...
+        self._closed = False
+
+    @workflow.run
+    async def run(self, input: HubInput) -> str:
+        await workflow.wait_condition(lambda: self._closed)
+        # The publisher publishes its own terminator into the stream
+        # before signaling close (see run_external_publisher.py).
+        # Hold the run open briefly so subscribers' final poll
+        # delivers any items still in the log.
+        await workflow.sleep(timedelta(milliseconds=500))
```
brianstrauch, line 35: Are there more dependable ways of checking if the log is not empty?

jssmith: Sometimes, but not always. We are not just checking that the log has been read, we need to know that the application is done reading before we end the workflow. Real applications may want to leave the workflow open for a longer period of time.

workflow_streams/activities/llm_activity.py
```diff
@@ -0,0 +1,74 @@
...
+    attempt's deltas, so consumers can reset their accumulated state
+    instead of concatenating the failed attempt's partial output with
+    the retried attempt's full output.
+
+    No ``force_flush=True``: the 200ms ``batch_interval`` is fast
+    enough for an interactive feel, and the WorkflowStreamClient's
+    ``__aexit__`` cancels a sleeping flusher cleanly. (The doc example
+    uses ``force_flush=True`` on the first delta; that path currently
+    wedges the activity's exit on a cancel-mid-flight bug — fix is
+    pending in ``temporalio.contrib.workflow_streams``.)
```
brianstrauch:
```suggestion
    ``__aexit__`` cancels a sleeping flusher cleanly.
```

README.md
```diff
@@ -79,6 +79,7 @@ Some examples require extra dependencies. See each sample's directory for specif
 * [patching](patching) - Alter workflows safely with `patch` and `deprecate_patch`.
 * [polling](polling) - Recommended implementation of an activity that needs to periodically poll an external resource waiting its successful completion.
 * [prometheus](prometheus) - Configure Prometheus metrics on clients/workers.
+* [workflow_streams](workflow_streams) - Workflow-hosted durable event stream via `temporalio.contrib.workflow_streams`. **Experimental — requires the [`contrib/pubsub` branch](https://github.com/temporalio/sdk-python/tree/contrib/pubsub) of sdk-python.**
```
brianstrauch:
```suggestion
* [workflow_streams](workflow_streams) - Workflow-hosted durable event stream via `temporalio.contrib.workflow_streams`. **Experimental**
```

brianstrauch (approved):

PR #301, opened 2026-04-30
Add openai_agents streaming sample

jssmith:
## Summary

Adds an `openai_agents/streaming/` sample demonstrating buffered token streaming for OpenAI Agents-backed workflows via `temporalio.contrib.workflow_streams` (experimental, [`contrib/pubsub`](https://github.com/temporalio/sdk-python/tree/contrib/pubsub) branch of sdk-python).

The OpenAI Agents plugin's `ModelActivityParameters` carries a `streaming_event_topic`; the model activity publishes raw `TResponseStreamEvent`s to that topic, batched over `streaming_event_batch_interval` (default 100ms). The workflow emits a sentinel on a `done` topic when `Runner.run_streamed` finishes; subscribers iterate `(events, done)` and break on the sentinel. `race_with_workflow` handles the case where the workflow fails before publishing the sentinel.

Two scenarios:

- `stream_text` — text-delta events from a simple haiku agent
- `stream_items` — agent-update / handoff / tool-call events across a multi-agent workflow with a joke-rating activity

This is the second half of #299. Independent of the workflow_streams basics (separate PR), though both target the same sdk-python contrib branch.

## Test plan

- [ ] Build sdk-python from `contrib/pubsub` and install into the samples uv environment
- [ ] `OPENAI_API_KEY=... uv run openai_agents/streaming/run_worker.py`
- [ ] `uv run openai_agents/streaming/run_stream_text_workflow.py` — verify text deltas print as small bursts
- [ ] `uv run openai_agents/streaming/run_stream_items_workflow.py` — verify agent-update / tool-call / message-output events render in order


brianstrauch:
Picked this up to get it over the line. Four commits pushed on top of yours (nothing rewritten, main merged in rather than rebased). The PR description above is now stale — summary of what changed:

### Updated for the released API

Workflow Streams and OpenAI Agents streaming both shipped in `temporalio` 1.30.0, so the `contrib/pubsub` install step is gone. Along with that:

- `streaming_event_topic` → `streaming_topic`, `streaming_event_batch_interval` → `streaming_batch_interval`.
- `subscribe()` without `result_type` now *decodes* payloads rather than handing back a raw `Payload`, so the `assert isinstance(item.data, Payload)` no longer holds. Subscribers pass `result_type=RawValue` and decode per topic, matching the merged `workflow_streams/` samples.
- Dropped `race_with_workflow` (all of `shared.py`'s 79 lines). Released `subscribe()` exits cleanly once the workflow reaches a terminal state, so the "subscriber blocked on its next poll forever" case it guarded against doesn't arise — break on the terminator, then `await handle.result()`. Confirmed by terminating a running workflow mid-subscribe: iterator exhausted cleanly, `handle.result()` raised `WorkflowFailureError`.
- Workflows now hold the run open briefly after publishing the terminator so a subscriber's last poll can drain the tail, same as `workflow_streams/workflows/llm_workflow.py`. The log lives in workflow memory, so it's gone the moment the run completes.

### `stream_items` needed a rethink

The streaming activity publishes **native OpenAI events**, not the agents-SDK `StreamEvent` wrappers — so the `agent_updated_stream_event` / `run_item_stream_event` / `tool_call_item` events that subscriber matched on never appear on that topic, and it would have rendered nothing but `=== Run starting ===`. Those higher-level events are built by the agents SDK inside the workflow.

So the workflow now publishes them itself, as a small serializable `ItemEvent` on its own `items` topic. It can't forward the SDK's own event types — `RunItemStreamEvent` carries the originating `Agent`, which holds tool callables. `stream_events()` resolves a turn at a time (one activity per model call), so the play-by-play still reaches the subscriber progressively rather than in one lump.

Same root cause in `stream_text`: it unwrapped a `raw_response_event` that isn't there. It now matches `ResponseTextDeltaEvent` directly.

### Tests

`tests/openai_agents/streaming_test.py` — first tests under that directory, shaped after `tests/strands_plugin/streaming_test.py`. A scripted streaming model (`_mock_model.py`) means no `OPENAI_API_KEY` is needed; unlike the strands and ADK mocks it needs no monkeypatching, since `OpenAIAgentsPlugin` accepts a `model_provider` directly. `test_stream_text` asserts the deltas arrive unwrapped and reassemble into exactly the workflow's return value (this is the regression test for the wrapped-vs-native bug); `test_stream_items` asserts the four event kinds arrive in order. Both subscribe the way the runner scripts do and assert the terminator is seen.

Green on the local dev server and under `--workflow-environment time-skipping`, ~3s per run. `poe lint` clean.

### Smaller things

Both workflow module docstrings sat *below* the imports, so they were dead string expressions rather than docstrings — moved to the top of the file. READMEs rewritten: branch-install notes dropped, the native-vs-wrapped event distinction documented, structure aligned with `openai_agents/basic/README.md`.

### Still worth doing before merge

I verified end-to-end against a scripted model, not a live one. A run with a real `OPENAI_API_KEY` would confirm the batched-burst cadence the README describes, which is the one claim a fake model can't substantiate.

openai_agents/README.md
```diff
@@ -36,3 +36,4 @@ Each directory contains a complete example with its own README for detailed inst
 - **[Customer Service](./customer_service/README.md)** - Interactive customer service agent with escalation capabilities, demonstrating conversational workflows.
 - **[Reasoning Content](./reasoning_content/README.md)** - Example of how to retrieve the thought process of reasoning models.
 - **[Financial Research Agent](./financial_research_agent/README.md)** - Multi-agent financial research system with planner, search, analyst, writer, and verifier agents collaborating.
+- **[Streaming](./streaming/README.md)** - `Runner.run_streamed` with buffered token streaming to external subscribers via `temporalio.contrib.workflow_streams`. **Experimental.**
```
DABH, line 39: Five places in openai_agents/ still say streaming isn't supported

The PR adds the feature but leaves the repo contradicting itself:

- openai_agents/basic/README.md:79 — - **Stream Items/Stream Text** - Streaming is not supported in Temporal OpenAI Agents SDK integration — under "Omitted Examples". This is the direct one: the PR is stream_items + stream_text.
- openai_agents/agent_patterns/README.md:97 — "Streaming Guardrails: Requires streaming capabilities which are not yet available in the Temporal integration"
- openai_agents/agent_patterns/README.md:43 — "(adapted for non-streaming)"
- openai_agents/handoffs/README.md:44 — "Message Filter Streaming: Streaming capabilities are not yet available in the Temporal integration"
- openai_agents/reasoning_content/README.md:35-37 — "## Note on Streaming … since Temporal workflows do not support streaming yet"

At minimum basic/README.md:79 must be redirected to ../streaming/. While rewriting it, state why these live in streaming/ rather than basic/: streaming_topic has to be set on ModelActivityParameters, so they can't share the basic worker. That preempts the obvious reviewer question and gives the line a natural replacement.

openai_agents/streaming/run_stream_text_workflow.py
```diff
@@ -0,0 +1,70 @@
...
+    async for item in stream.subscribe(
+        [TOPIC_EVENTS, TOPIC_DONE], result_type=RawValue
+    ):
+        if item.topic == TOPIC_DONE:
+            break
+        # Subscribers receive native OpenAI events, not the agents-SDK
+        # StreamEvent wrappers that stream_events() yields in the workflow.
+        event: Any = converter.from_payload(item.data.payload, EVENT_TYPE)
+        if isinstance(event, ResponseTextDeltaEvent):
+            print(event.delta, end="", flush=True)
```
DABH, line 90: A model-activity retry produces visibly garbled subscriber output — reproduced

Scripted a model that streams 4 deltas, raises, then succeeds on retry. Subscriber is the exact run_stream_text_workflow.py:53-62 shape:

subscriber reassembled text:
```
  'Why did the developer go broke? He used up all hWhy did the developer go broke? He used up all his cache. Why do programmers prefer dark mode? Light attracts bugs.'
workflow returned:
  'Why did the developer go broke? He used up all his cache. Why do programmers prefer dark mode? Light attracts bugs.'
MATCH: False
```

The failed attempt's 48-character prefix is rendered verbatim, then the whole answer restarts. The workflow's return value stays correct (stream_events() only sees the successful attempt), so only the streamed view is corrupt — the harder failure to notice.

It isn't an in-flight race. _invoke_model_activity.py:402 wraps the model loop in async with stream:, and WorkflowStreamClient.__aexit__ (_client.py:226-241) unconditionally drains while self._pending is not None or self._buffer. Everything produced before the raise is durably signalled before the activity reports failure. And each attempt gets a fresh publisher_id (_client.py:128, uuid.uuid4().hex[:16]), so the dedup logic deliberately does not suppress it.

README.md:116-122 describes this in prose, but the render site has no mitigation and no comment, and the mitigation it points at doesn't exist:

> The workflow_streams module documentation (…) covers the trade and the conventional RETRY event pattern for surfacing it to consumers.

That pattern is something you hand-roll in your own activity — workflow_streams/activities/llm_activity.py:48-52 does exactly that with activity.info().attempt. The plugin's streaming activity has no such hook and emits no retry marker (grepped _invoke_model_activity.py). Worse, the linked page is a 33-line stub; on sdk-python@origin/main the only retry-adjacent line in it says the opposite:

```
18: a usable stream: batching to amortize per-event overhead, deduplication
19: for exactly-once delivery, topic filtering, and continue-as-new helpers
```

Suggested fix: say plainly that the plugin does not emit a retry marker today, drop the dead link, and give subscribers the signal that is available — a real OpenAI stream restarts sequence_number at 0 and re-emits response.created on each attempt, so a delta-reassembling subscriber can reset on that. A three-line if in run_stream_text_workflow.py would turn the sample's worst failure mode into something it teaches.

openai_agents/streaming/run_worker.py
```diff
@@ -0,0 +1,54 @@
...
+                model_params=ModelActivityParameters(
+                    # The streaming activity heartbeats on a background task,
+                    # so a heartbeat_timeout well under start_to_close_timeout
+                    # detects a stuck model call early.
+                    heartbeat_timeout=timedelta(seconds=10),
+                    start_to_close_timeout=timedelta(minutes=5),
+                    # Required for Runner.run_streamed: the topic the streaming
+                    # activity publishes raw model events to.
+                    streaming_topic=TOPIC_EVENTS,
+                ),
```
DABH, line 39: run_worker.py:30-39 never sets streaming_batch_interval. The README's central paragraph (:20-26) is about buffering and tuning that knob, but the code only sets streaming_topic and takes the default. Setting it explicitly with a one-line comment makes the sample self-describing.

openai_agents/streaming/README.md
```diff
@@ -0,0 +1,122 @@
...
+small bursts rather than glyph-by-glyph. Lower the interval for smoother
+output at the cost of more signals.
+
+Two things to know before reading the samples:
+
+* `streaming_topic` is **required** for `Runner.run_streamed`. If it is unset,
+  `run_streamed` raises before scheduling any activity.
+* The workflow must host a `WorkflowStream`, constructed in `@workflow.init` so
+  the publish-signal handler is registered before the activity publishes.
+  Without one, the publishes are unhandled and silently dropped.
```
DABH: README.md:31-34 says the stream is "constructed in @workflow.init so the publish-signal handler is registered before the activity publishes. Without one, the publishes are unhandled and silently dropped." Right code, wrong reasons: _stream.py:127-133 enforces __init__ construction (it raises otherwise), and unhandled signals aren't dropped — the SDK buffers them for later-registered dynamic handlers (_workflow_instance.py:387,1490-1498).

openai_agents/streaming/workflows/stream_text_workflow.py
```diff
@@ -0,0 +1,79 @@
...
+                deltas += 1
+        workflow.logger.info("collected %d delta events", deltas)
+
+        # In-band terminator so the subscriber can stop without racing the
+        # workflow's completion, then a brief pause to let its next poll
+        # deliver the tail of the stream — the log lives in workflow memory
+        # and is gone once this run completes.
+        self.done.publish(True)
+        await workflow.sleep(DRAIN_INTERVAL)
+        return result.final_output
```
DABH: stream_text_workflow.py:79 and stream_items_workflow.py:94 return result.final_output (Any) from -> str. final_output is None when a run produces no message output. mypy passes only because the field is Any.

DABH (changes requested):

openai_agents/streaming/run_worker.py
brianstrauch: Intentionally left it out since most use cases would take the default

DABH (approved):
LGTM

tconley1428 (approved):

PR #307, opened 2026-05-13
External storage and codec server sample

jmaeagle99:
## What was changed

Add sample for external storage with codec server that supports multiple namespaces and downloads.

## Why?

Provide a baseline reference for how to use external storage with the S3 driver and an updated codec server.

## Checklist

1. How was this tested: Locally ran; investigating possibility of adding tests
1. Any docs updates needed? No


cconstable (approved):
👍 Very neat. Learned about `X-Namespace` and `preserveStorageRefs` from reviewing this.

PR #313, opened 2026-05-21
fix(external-storage): correct the reference check

jmaeagle99:
## What was changed

Fix the storage reference check in the external-storage sample.

### DIFF
--- a/external_storage/handler.py
+++ b/external_storage/handler.py
@@ -22,17 +22,20 @@ from typing import List, Optional
 from aiohttp import web
 from google.protobuf import json_format
 from temporalio.api.common.v1 import Payload, Payloads
+from temporalio.api.sdk.v1 import ExternalStorageReference
 from temporalio.converter import ExternalStorage, PayloadCodec
+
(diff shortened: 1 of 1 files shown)


cconstable (approved):

external_storage/handler.py
```diff
@@ -22,17 +22,20 @@
...
-    repeated field is the marker the SDK uses to recognize it on the way back.
+def _is_storage_reference(payload: Payload) -> bool:
+    """A payload is an external-storage reference iff its metadata identifies
+    it as a serialized :class:`ExternalStorageReference` claim check.
     """
-    return len(payload.external_payloads) > 0
+    return (
+        payload.metadata.get("encoding") == b"json/protobuf"
+        and payload.metadata.get("messageType") == _REFERENCE_MESSAGE_TYPE
+    )
```
cconstable, line 38: I believe there is also a legacy encoding `json/external-storage-reference`. Does that need to be handled?

jmaeagle99: Probably not and we're likely to remove it soon. We can update to include it if we get feedback.

PR #315, opened 2026-06-04
Add LangGraph workflow streams sample

brianstrauch:
## What

Adds a Graph-API sample demonstrating the LangGraph plugin's recently-added [Workflow Streams](https://github.com/temporalio/sdk-python/tree/main/temporalio/contrib/workflow_streams) support, under `langgraph_plugin/graph_api/streaming/`.

It combines both streaming patterns in one workflow:
- **Node token streaming** — the `write_story` node emits live tokens via LangGraph's `get_stream_writer()`, routed onto the `tokens` topic by the plugin's `streaming_topic`.
- **Workflow-side `astream` publish** — the workflow drives the graph with `app.astream(...)` and publishes each node-completion chunk onto a `progress` topic it owns via `WorkflowStream.topic().publish()`.

A single client subscribes to all topics with `WorkflowStreamClient` and demultiplexes on `item.topic`. The workflow waits on an `ack_stream` signal before returning so the stream survives until the client finishes consuming it.

## Dependency changes

- Bumps the `langgraph` group to `temporalio>=1.28.0` (where workflow streams ship).
- Drops the now-obsolete `langsmith<0.7.34` uv constraint — 1.28.0 requires `langsmith>=0.7.34`, and the pin was specific to the 1.27.2 langsmith patch. Existing langgraph tests still pass on the resulting langsmith 0.8.9.

Also adds `langgraph_plugin/` code owners.

## Testing

- Added `tests/langgraph_plugin/streaming_test.py`; the full `tests/langgraph_plugin/` suite passes (12 passed) on temporalio 1.28.0 / langsmith 0.8.9.
- `ruff` format/import checks and `mypy` are clean for the new files.


langgraph_plugin/README.md
DABH, line 1: Every other graph_api/ sample has its own README.md; graph_api/streaming/ is the only one without. The module docstring in workflow.py is excellent, but the convention is a standalone README with run instructions and a "what this demonstrates" blurb. Can you please add a langgraph_plugin/graph_api/streaming/README.md?

langgraph_plugin/graph_api/streaming/run_workflow.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Subscribe to all topics on the workflow's stream and demultiplex on topic.
+    ws = WorkflowStreamClient.create(client, handle.id)
+    async for item in ws.subscribe(
+        from_offset=0,
+        result_type=dict,
+        poll_cooldown=timedelta(milliseconds=50),
+    ):
+        if item.topic == "tokens":
+            print(item.data["token"], end="", flush=True)
```
DABH, line 38: The LangGraph plugin README states this in bold: "Streaming has *at-least-once* delivery per activity attempt. When an activity-wrapped node retries … the user function re-runs from scratch and re-publishes its writes … Subscribers should be ready to see duplicates and recover idempotently (e.g. dedupe on a sequence id …)."

The sample does the opposite of modeling that: the client concatenates tokens (here on this line) and the test asserts exact reassembly (streaming_test.py:57-58). If `write_story` ever retried (transient failure, worker restart), tokens would duplicate and you'd get garbled output and a failing assertion.  (I suspect it passes in CI only because the in-process fake never triggers a retry.) So the sample implicitly teaches exactly-once for a stream the plugin explicitly documents as at-least-once.

Could you document in the README for this example that that's what's going on (and noting that consumers should dedup), or even better, maybe update this example to demonstrate the documented idempotent pattern (a sequence id per chunk)?

DABH (changes requested):

DABH (approved):

PR #321, opened 2026-06-25
AI-284: Add snipsync markers and trim comments in ADK samples

xumaple:
Adds SNIPSTART markers around the Worker/Client setup and Workflow starter so the documentation integration page (temporalio/documentation#4759) can pull them via snipsync, and trims the sample comments to short, why-only notes.


DABH (approved):
These all look like reasonable condensations of the comments, thanks for making this pass through

PR #330, opened 2026-07-24
Add Google ADK chatbot sample

xumaple:
Google ADK chatbot sample using updates

### DIFF
--- a/google_adk_agents/chatbot/workflows/chatbot_workflow.py
+++ b/google_adk_agents/chatbot/workflows/chatbot_workflow.py
@@ -1,0 +1,70 @@
+import asyncio
+
+from google.adk import Agent
+from google.adk.runners import InMemoryRunner
+from google.adk.utils.context_utils import Aclosing
+from google.genai import types
+from temporalio import workflow
+from temporalio.contrib.google_adk_agents import TemporalModel
+
+
+# @@@SNIPSTART google-adk-agents-chatbot-agent-workflow
+@workflow.defn
+class ChatbotAgentWorkflow:
+    def __init__(self) -> None:
+        self._ready = False
+        self._runner: InMemoryRunner | None = None
+        self._session_id: str | None = None
+
+    @workflow.run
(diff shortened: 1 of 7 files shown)


google_adk_agents/chatbot/workflows/chatbot_workflow.py
```diff
@@ -0,0 +1,70 @@
...
+        async with Aclosing(
+            self._runner.run_async(
+                user_id="user",
+                session_id=self._session_id,
+                new_message=types.Content(
+                    role="user", parts=[types.Part(text=message)]
+                ),
+            )
+        ) as agen:
+            async for event in agen:
```
brianstrauch: This looks like you spelled agent wrong. Maybe it stands for "async generation"?

google_adk_agents/chatbot/workflows/chatbot_workflow.py
```diff
@@ -0,0 +1,70 @@
...
+                if event.content and event.content.parts:
+                    for part in event.content.parts:
+                        if part.text:
+                            final_text = part.text
+
+        return final_text
+
+    @message.validator
+    def validate_message(self, message: str) -> None:
+        pass
```
brianstrauch, line 68: This doesn't appear to be used, and it has an empty body

xumaple: That's intentional to demonstrate it. You think unnecessary? I added a comment

google_adk_agents/chatbot/workflows/chatbot_workflow.py
xumaple: renamed

google_adk_agents/chatbot/workflows/chatbot_workflow.py
brianstrauch: I think I'm missing context as to why it's needed. Will teach myself!

brianstrauch (approved):

PR #331, opened 2026-07-28
Add LangFuse tracing samples

DABH:
Adds samples of using our OTel plugin to send traces to LangFuse from Temporal workflows.


.github/CODEOWNERS
```diff
@@ -1,5 +1,4 @@
 * @temporalio/sdk
-/langgraph_plugin/ @temporalio/sdk @temporalio/ai-sdk
```
DABH, line 2: This is a duplicate of L19 down below, can be safely removed

xumaple (commented):
Overall looks good, there were a few comments-related things that claude pulled up:

- starter.py:51 + README.md:131 — "unique workflow ID gives one Langfuse trace per run" is causally wrong. The trace ID comes from the starter's root span (starter.py:66), fresh every run regardless of workflow ID; what the unique ID buys is a unique langfuse.session.id, which is what --workflow-id resolution depends on.
- verify_trace.py:6 references a "replay-stress test" that doesn't exist — --replay-stress is a worker CLI flag. Surfaces in --help via description=__doc__.
- README.md:76-77 — "the identical tree" is wrong (each run is a new trace; the shape is identical), and it tells you to rerun verify_trace.py after each experiment, but neither following command block contains that invocation.
- docker-compose.yml:6-8 — "upstream binds them to 127.0.0.1" is false for MinIO (upstream publishes 9090:9000 on all interfaces), and the list omits langfuse-worker. Related: docker-compose.yml:57,65 still point Langfuse's S3 media/export external endpoints at localhost:9090, a port this file deliberately stopped publishing — any presigned media URL is a dead link (unreachable in this text-only demo).
- .env.example omits LANGFUSE_DEMO_USER, which starter.py:73 reads and verify_trace.py:208 asserts on (PLAUSIBLE — the file otherwise enumerates the full surface).
- README.md:86-87 — "kill the worker while the workflow is parked" is imprecise; early in the 20s window the workflow is still running classify_ticket, so you'd exercise activity retry instead (PLAUSIBLE).
- "Idempotent" for the LANGFUSE_INIT_* provisioning couldn't be verified offline (PLAUSIBLE).

DABH:
Thanks @xumaple — all good catches. Addressed in 42cd6e6:

- **Trace/session causality** (starter.py + README): reworded both. The Langfuse trace is keyed by the starter's root span, which is new on every run; what the fresh workflow ID buys is a unique `langfuse.session.id` (which `verify_trace --workflow-id` resolution depends on) and no Temporal workflow-ID conflicts.
- **verify_trace.py docstring**: now says "running the worker with `--replay-stress`" instead of referencing a nonexistent test.
- **README replay experiments**: "identical tree" → "same tree shape", added the `verify_trace` invocation to both command blocks, and made the worker-restart timing precise (kill the worker after the triage activities finish, while the workflow durably awaits approval — killing earlier exercises activity retry instead, which is a different demo).
- **docker-compose**: MinIO's S3 endpoint is now published loopback-only (`127.0.0.1:9090:9000`), matching the `LANGFUSE_S3_MEDIA_UPLOAD_ENDPOINT`/`LANGFUSE_S3_BATCH_EXPORT_EXTERNAL_ENDPOINT` defaults so presigned URLs actually resolve (verified `/minio/health/live` returns 200 from the host). The header comment is corrected too — upstream publishes MinIO on all interfaces (not 127.0.0.1), and langfuse-worker is now included in the "stays on the compose network" list.
- **`LANGFUSE_DEMO_USER`**: added to `.env.example`.
- **"Idempotent"**: reworded to what Langfuse documents — provisioning is skipped when the org/project/user already exist, so restarts are safe (consistent with what we've observed across many restarts of this stack).

Re-ran ruff/mypy and the sample tests after the changes.

xumaple (approved):
LGTM

tconley1428 (approved):

PR #351, opened 2026-08-18
Add a sandbox sample and snipsync markers for the OpenAI Agents docs guide

brianstrauch:
## What changed

**New `openai_agents/sandbox` sample.** Covers `SandboxAgent` through the plugin: `SandboxClientProvider` on the worker, `temporal_sandbox_client()` in the workflow, and a run that writes a file in the sandbox, reads it back, and reports what it found.

It uses `UnixLocalSandboxClient`, so it runs with only `OPENAI_API_KEY` and no Daytona/E2B account — at the cost of giving the agent a shell on the worker host, which the sample README calls out and points at the remote clients for production. Worth knowing: that class lives in `agents.sandbox.sandboxes.unix_local`, not the `agents.extensions.sandbox.unix_local` path the contrib README shows (fixed in temporalio/sdk-python#1764).

**Snipsync markers** on the 16 existing samples the docs guide walks through.

## Why?

Python is currently the only Temporal OpenAI Agents integration with no guide on docs.temporal.io — the SDK index and the integrations grid both send readers to the contrib README on GitHub. The companion docs PR adds that guide, and it needs code to pull from here rather than hand-maintaining inline copies. Sandbox is the one section with no sample to pull from, hence the new one.

Two details worth flagging for review:

- Markers are scoped to exactly the code the guide shows, so no docs block needs `selectedLines` and none renders a leading `# ...` elision.
- Markers that wrap indented code sit at that code's indent level rather than column 0, keeping snipsync's dedenting working — the same fix as #346.

## Checklist

1. Closes 

2. How was this tested:

The sandbox sample was run end to end against `temporal server start-dev`. With no OpenAI key available, the model was driven by the plugin's own `temporalio.contrib.openai_agents.testing.TestModel` scripting one `exec_command` tool call; the sandbox itself was real. The workflow returned its output and history confirms the sandbox activities executed:

```
invoke_model_activity
local-sandbox_client_create
local-sandbox_session_start
local-sandbox_session_running
local-sandbox_session_pty_exec_start
local-sandbox_session_stop
local-sandbox_session_shutdown
local-sandbox_client_delete
```

Also confirmed the worker registers all 16 activities under the `local-` provider prefix, and that it boots and polls the task queue.

`ruff check --select I`, `ruff format --check`, and `mypy --check-untyped-defs --namespace-packages` all pass across `openai_agents/` (120 files).

Snippet resolution was verified by pointing a local `snipsync.config.yaml` at this working tree and running `yarn snipsync` from the documentation repo: all 18 IDs resolve and every excerpt dedents to column 0.

### DIFF
--- a/openai_agents/README.md
+++ b/openai_agents/README.md
@@ -39,4 +39,5 @@ Each directory contains a complete example with its own README for detailed inst
 - **[Customer Service](./customer_service/README.md)** - Interactive customer service agent with escalation capabilities, demonstrating conversational workflows.
 - **[Reasoning Content](./reasoning_content/README.md)** - Example of how to retrieve the thought process of reasoning models.
 - **[Financial Research Agent](./financial_research_agent/README.md)** - Multi-agent financial research system with planner, search, analyst, writer, and verifier agents collaborating.
+- **[Sandbox](./sandbox/README.md)** - `SandboxAgent` with a shell and filesystem, where every sandbox operation runs as a Temporal activity. **Pre-release.**
 - **[Streaming](./streaming/README.md)** - `Runner.run_streamed` with buffered token streaming to external subscribers via `temporalio.contrib.workflow_streams`. **Experimental.**
--- a/openai_agents/README.md
+++ b/openai_agents/README.md
@@ -39,5 +39,5 @@ Each directory contains a complete example with its own README for detailed inst
 - **[Customer Service](./customer_service/README.md)** - Interactive customer service agent with escalation capabilities, demonstrating conversational workflows.
 - **[Reasoning Content](./reasoning_content/README.md)** - Example of how to retrieve the thought process of reasoning models.
 - **[Financial Research Agent](./financial_research_agent/README.md)** - Multi-agent financial research system with planner, search, analyst, writer, and verifier agents collaborating.
-- **[Sandbox](./sandbox/README.md)** - `SandboxAgent` with a shell and filesystem, where every sandbox operation runs as a Temporal activity. **Experimental.**
+- **[Sandbox](./sandbox/README.md)** - `SandboxAgent` with a shell and filesystem, where every sandbox operation runs as a Temporal activity. **Pre-release.**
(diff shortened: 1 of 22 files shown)


openai_agents/README.md
```diff
@@ -39,4 +39,5 @@ Each directory contains a complete example with its own README for detailed inst
 - **[Customer Service](./customer_service/README.md)** - Interactive customer service agent with escalation capabilities, demonstrating conversational workflows.
 - **[Reasoning Content](./reasoning_content/README.md)** - Example of how to retrieve the thought process of reasoning models.
 - **[Financial Research Agent](./financial_research_agent/README.md)** - Multi-agent financial research system with planner, search, analyst, writer, and verifier agents collaborating.
+- **[Sandbox](./sandbox/README.md)** - `SandboxAgent` with a shell and filesystem, where every sandbox operation runs as a Temporal activity. **Pre-release.**
```
brianstrauch:
```suggestion
- **[Sandbox](./sandbox/README.md)** - `SandboxAgent` with a shell and filesystem, where every sandbox operation runs as a Temporal activity. **Experimental.**
```

openai_agents/README.md
```diff
@@ -39,4 +39,5 @@ Each directory contains a complete example with its own README for detailed inst
 - **[Customer Service](./customer_service/README.md)** - Interactive customer service agent with escalation capabilities, demonstrating conversational workflows.
 - **[Reasoning Content](./reasoning_content/README.md)** - Example of how to retrieve the thought process of reasoning models.
 - **[Financial Research Agent](./financial_research_agent/README.md)** - Multi-agent financial research system with planner, search, analyst, writer, and verifier agents collaborating.
+- **[Sandbox](./sandbox/README.md)** - `SandboxAgent` with a shell and filesystem, where every sandbox operation runs as a Temporal activity. **Experimental.**
```
brianstrauch:
```suggestion
- **[Sandbox](./sandbox/README.md)** - `SandboxAgent` with a shell and filesystem, where every sandbox operation runs as a Temporal activity. **Pre-release.**
```

xumaple (approved):

PR #352, opened 2026-08-18
Nexus query sdk ergonomics

JoshuaFrenchwood:
## What was changed

  - Use TemporalNexusClient for queries, updates, and signals
  - Return synchronous TemporalOperationResult values

## Checklist

1. Closes 

2. How was this tested:

- Ran the samples to test the result

3. Any docs updates needed?


Evanthx (approved):
I think it looks pretty good!

PR #359, opened 2026-09-04
Updating Nexus Messaging start_workflow example

JoshuaFrenchwood:
## What was changed

Small Nexus Messaging update to use temporal_operation decorator for starting workflow. instead of workflow_run_operation

## Checklist

1. Closes 

2. How was this tested:

Ran the sample and tested the result

3. Any docs updates needed?


atol (approved):

Evanthx (approved):
LGTM

PR #360, opened 2026-09-08
Introduce AgentCore strands-agent sample

justinschoeff:
## What was changed
This sample includes a strands agent loop aimed to run on AWS AgentCore,
invoked by Temporal serverless workers. This utilizes the AWS AgentCore
CLI and related configuration for serverless workers.

## Why?
AWS AgentCore has a unique interface and is a key service being marketed by AWS. This is in preparation for launch of serverless workers supporting AgentCore.

## Checklist

1. Closes 

2. How was this tested:
Ran the commands in readme, verified it created runtime, started the workflow and watched the worker in AgentCore runtime trigger activities in CW logs and in workflow history.

3. Any docs updates needed?
@lennessyy is working on docs


bedrock_agentcore/strands-agent/README.md
```diff
@@ -0,0 +1,193 @@
...
+*and* its endpoints. Give the stack's `RoleARN` output, and the endpoint, back to Temporal Cloud.
+
+### 5. Set up the Worker Deployment Version
+
+The Worker always registers the Worker Deployment Version named by
+`TEMPORAL_DEPLOYMENT_NAME` / `TEMPORAL_BUILD_ID`, and pins Workflows to it. You
+must set that version as current, or nothing will be routed to the Worker:
+
+```bash
+
```
brianstrauch, line 148:
```suggestion
```

pyproject.toml
```diff
@@ -132,12 +132,12 @@ log_cli_format = "%(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(linen
...
 [tool.ruff]
 target-version = "py310"
-extend-exclude = ["lambda_worker"]
+extend-exclude = ["bedrock_agentcore", "lambda_worker"]
 
 [tool.mypy]
 ignore_missing_imports = true
 namespace_packages = true
-exclude = ["lambda_worker/"]
+exclude = ["bedrock_agentcore/", "lambda_worker/"]
```
brianstrauch, line 140: What's behind the decision not to lint/typecheck?

README.md
```diff
@@ -62,6 +62,7 @@ Some examples require extra dependencies. See each sample's directory for specif
 * [activity_worker](activity_worker) - Use Python activities from a workflow in another language.
 * [batch_sliding_window](batch_sliding_window) - Batch processing with a sliding window of child workflows.
 * [bedrock](bedrock) - Orchestrate a chatbot with Amazon Bedrock.
+* [bedrock_agentcore/strands-agent](bedrock_agentcore/strands-agent) - Run a AWS Strands Agent with Temporal Plugin on AgentCore Worker.
```
brianstrauch, line 65: Use underscores for directory name consistency?
```suggestion
* [bedrock_agentcore/strands_agent](bedrock_agentcore/strands_agent) - Run a AWS Strands Agent with Temporal Plugin on AgentCore Worker.
```

brianstrauch: Why do we have a strands-agent subdirectory in the first place? Planning on demoing other agent frameworks with AgentCore?

bedrock_agentcore/strands-agent/agentcore_worker.py
```diff
@@ -0,0 +1,145 @@
...
+            self._tracker.inflight -= 1
+            self._tracker.changed.set()
+# @@@SNIPEND
+
+
+# @@@SNIPSTART python-agentcore-runtime-handler
+@app.entrypoint
+@app.async_task  # keeps /ping on "HealthyBusy" until this returns
+async def invoke(payload: dict) -> dict:
+    """Poll until idle, then drain. The payload is unused: every call is a new session and new worker."""
```
brianstrauch, line 105: Would it make sense to launch the Worker in a background task/thread, register it with add_async_task, and immediately return an acknowledgement?

bedrock_agentcore/strands-agent/agentcore_worker.py
```diff
@@ -0,0 +1,145 @@
...
+        worker = Worker(
+            client,
+            task_queue=task_queue,
+            workflows=[workflows.StrandsAgentWorkflow],
+            activities=[execute_code],
+            activity_executor=activity_executor,
+            interceptors=[tracker],
+            deployment_config=WorkerDeploymentConfig(
+                version=WorkerDeploymentVersion(
+                    deployment_name=DEPLOYMENT_NAME, build_id=BUILD_ID
```
brianstrauch, line 130: Consider passing deployment name and build ID in the payload. Using environment variables as a second source of truth could cause version drift

bedrock_agentcore/strands_agent/agentcore_worker.py
```diff
@@ -0,0 +1,145 @@
...
+        worker = Worker(
+            client,
+            task_queue=task_queue,
+            workflows=[workflows.StrandsAgentWorkflow],
+            activities=[execute_code],
+            activity_executor=activity_executor,
+            interceptors=[tracker],
+            deployment_config=WorkerDeploymentConfig(
+                version=WorkerDeploymentVersion(
+                    deployment_name=DEPLOYMENT_NAME, build_id=BUILD_ID
```
justinschoeff: We have been discussing this, but its a double-edged swored imo. The version should be strongly tied to the code. As is, blindly running a worker with an externally provided version is asking for trouble. With AgentCore the ENV variables are tied to the version, so it's a bit safer. I think static in code is probably safest, but passing in payload is easiest to use / dev friendly cause it allows things to just work.

README.md
justinschoeff: I also have a helloworld-container version that builds a Docker image and has the http server in example. I'm not sure it's useful for python as it would be for other languages that don't support AgentCores CodeZip sdk. I was going to PR right after this, but I may not. I do think we'll want other examples at some point though. Open to moving it up a level.

bedrock_agentcore/strands-agent/agentcore_worker.py
justinschoeff: Updated

pyproject.toml
justinschoeff: removed ruff. the mypy is becasue both have activities in their root and it complains about conflicting modules.

brianstrauch: Would the fix be to split the activities into a separate file?

README.md
brianstrauch: Seems reasonable to move it up a level until we add a second sample

bedrock_agentcore/strands_agent/agentcore_worker.py
brianstrauch: Yep, there are pros and cons with both approaches, so let's just make sure documentation is solid for whichever one we choose!

brianstrauch (approved):

lennessyy:
Sorry - was trying to test the docs using the sample but got blocked on Anthropic calls. I asked Codex to change the model to nova lite and it went rogue and pushed changes here hence the push and revert. Changes are already reverted and I only added snipsync comments

bedrock_agentcore/strands_agent/pyproject.toml
```diff
@@ -0,0 +1,54 @@
+[project]
```
tconley1428, line 1: Why are we making nested pyprojects?

justinschoeff: This alongside the lambda_worker are for serverless workers and the sample itself includes bundling and deploying the code. Here, its in the create-runtime.sh which bundles the sample code and deploys to AWS AgentCore.

There's also some version conflicts around bedrock_agentcore here and aioboto3 in the root.

tconley1428 (approved):

PR #366, opened 2026-09-11
Add OpenRouter samples: prompt batch and budget gate

DABH:
Adds an `openrouter` sample with two scenarios, plus OpenRouter as a model provider for the OpenAI Agents SDK sample.

**prompt_batch** fans one Activity out per prompt with OpenRouter's Auto Router and returns answer, model, and reported cost per prompt. The Activity uses the `openai` client pointed at OpenRouter with `max_retries=0`, classifies 4xx as non-retryable and 408/429/5xx as retryable, passes `Retry-After` through as `next_retry_delay`, heartbeats, and sends `X-OpenRouter-Cache: true` so a retried identical request is served from OpenRouter's cache at $0. A `--fail-once` flag demonstrates that.

**budget_gate** is the same batch, but it pauses on a soft budget or on OpenRouter's 402 and resumes on a `raise_budget` Update, with a `spend_report` Query.

Tests use `httpx.MockTransport` and a fake Activity; no API key needed. Verified live against OpenRouter, including a cache hit at $0 on retry.

Docs page: temporalio/documentation#5307. TypeScript port: temporalio/samples-typescript#520.


openai_agents/model_providers/workflows/openrouter_workflow.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import annotations
+
+from agents import Agent, Runner, function_tool
+from temporalio import workflow
+
+# Any OpenRouter model slug works here. A fixed, tool-capable model keeps the
+# sample reproducible; swap in "openrouter/auto" to let OpenRouter choose.
+OPENROUTER_MODEL = "openai/gpt-4o-mini"
```
patbeqo, line 8: Should we default to `openrouter/auto` since the README says `and picks providers and models per request` ?

